PR #10, opened 2023-07-27
Better ios integration

yssf-io:
* draft swift to send more passport data
* same button as android to scan


0xturboblitz:
ended up copying changes manually to https://github.com/zk-passport/proof-of-passport/tree/ios

PR #43, opened 2024-01-03
merkle tree inclusion proof in /contract and /circuit

yssf-io:


0xturboblitz:
It's all done now except the mysterious proof generation bug on android

0xturboblitz:
- Switching to mopro-android solves the mysterious bug aforementioned.
- This PR temporarily switches back to local zkey loading instead of dynamic. This will be fixed soon when merging the dylib branch.
- TODO: handle case in which signature is valid but public key is not in the tree

PR #71, opened 2024-03-04
Integrate circom-witness-rs

0xturboblitz:
- importing circom-witness-rs
- have to implement new operations to make graph.bin gen work with our circuit, c.f. https://github.com/philsippl/circom-witness-rs/issues/8
- deleted some dylib functions

This would:
1. Make witness gen way faster
2. Avoid the .wasm, .dylib and .framework issues on iOS

Example use can be found [here](https://github.com/worldcoin/semaphore-rs/pull/58)


0xturboblitz:
Closing for now as the polygon witnesscalc is pretty fast so not planning on switching soon

Issue #41, opened 2023-12-31
Add range check to circuit


0xturboblitz:
It would be cool to be able to check if a user is older than 18 or 21, instead of having him reveal its whole date of birth.

remicolin:
Working on it on the branch circuits

0xturboblitz:
Merged in https://github.com/zk-passport/proof-of-passport/pull/75

PR #75, opened 2024-03-15
Age range check

0xturboblitz:
- Adding age range check to circuits and app


remicolin:
This PR adds a range check for user age, closing #41

Two inputs have been added to the main circuit:

- current_date[6]: the current date in YYMMDD format represented as an array of numbers, e.g. [2, 4, 0, 1, 0, 1].
- majority[2]: the age on which the circuit will perform the range check, represented in ASCII format, e.g. [49, 56] for 18.

The SBT has been extended with one attribute: older_than, which will have majority[2] as its value.

PR #84, opened 2024-04-06
Moving to PolygonID's stack on Android

0xturboblitz:
This is preliminary work on moving to PolygonID's stack on Android.
This is not necessary right now as mopro works well on Android for us, and the blocker was doing a release on iOS.

- I fixed `fr_raw_generic.cpp` dirtily by typecasting, which might have broken witness calculation.
=> As suggested by their team, look at [this build script](https://github.com/0xPolygonID/witnesscalc/blob/main/.github/workflows/build.yml) to understand why it doesn't build well on my side.

- At the zkey loading step, it says:
`
Could not invoke Prover.runProveAction
java.lang.OutOfMemoryError: Failed to allocate a 268435472 byte allocation with 25165824 free bytes and 115MB until OOM, target footprint 172835488, growth limit 268435456
`
=> Looks like `android:largeHeap="true"` does not solve, turns into:
`
Caused by: java.lang.OutOfMemoryError: Failed to allocate a 536870928 byte allocation with 25165824 free bytes and 242MB until OOM, target footprint 307414080, growth limit 536870912
`
=> no trivial solution but look at the react-native-rapidsnark update they pushed

- It looks from [here](https://github.com/rarimo/FreedomToolAndroid/tree/main/app/src/main/cpp) that we should create a cpp wrapper and build it with CMake. This is necessary to have function names compliant with JNI and also typecasting correctly. Right now I have trouble doing that.
- Typing of inputs in `ProverModule.kt` probably incorrect


0xturboblitz:
fixed 1 using the team's fix https://github.com/0xPolygonID/witnesscalc/issues/25, 2 using `groth16ProveWithZKeyFilePath` from [react-native-rapidsnark](https://github.com/iden3/react-native-rapidsnark), and 3 et 4 were fine

Issue #109, opened 2024-05-23
Implement MGF1 in circom


0xturboblitz:
Implement MGF1 in circom, which will be a key part of the verification of RSA-PSS in circom, c.f. https://github.com/zk-passport/proof-of-passport/issues/38.

I would take inspiration from [this](https://github.com/shigeki/ohtsu_rsa_pss_js/blob/41f98614e9080080103db6b23e79cf6d8b78f794/rsa_pss.js#L114).

Prabhat1308:
Hi. I have some experience in Circom and would like to take up this issue !

0xturboblitz:
Cool please explain how you would tackle it

seshanthS:
Hey @0xturboblitz, I would like to work on this issue. 

I am a Solidity Dev, with some Circom experience. I have used it in few hobby projects before.

I found this spec (https://datatracker.ietf.org/doc/html/rfc8017#appendix-B.2.1) and will use it to tackle this issue

remicolin:
Nice @Prabhat1308 @seshanthS,
Can you take a look at the circuits repo and give a detailed plan on how you will tackle the issue ?
If you want to work on circom circuits there is also other signatures algorithm to implement.

0xturboblitz:
Closing, done in https://github.com/zk-passport/proof-of-passport/pull/115

PR #115, opened 2024-05-28
feat: Implement MGF1

seshanthS:
Implements #109 

Tests:
from `proof-of-passport/circuits` directory,

`yarn ts-mocha test/Mgf1_sha256.test.ts `

### DIFF
--- a/circuits/circuits/tests/mgf1Sha256/Mgf1Sha256_1ByteMask_tester.circom
+++ b/circuits/circuits/tests/mgf1Sha256/Mgf1Sha256_1ByteMask_tester.circom
@@ -1,0 +1,31 @@
+pragma circom 2.1.5;
+
+include "../../utils/Mgf1Sha256.circom";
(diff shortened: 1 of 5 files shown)


remicolin:
Hi @seshanthS 
MGF1 has to be generic and work with all SHA algorithms.
Please don't hardcode the shaLen in MGF1´s circuit and implement tests with other SHA functions.

0xturboblitz:
Hi @seshanthS 

This is really cool!

I just added the mock passport data for rsassaPss here: https://github.com/zk-passport/proof-of-passport/blob/57ec1c3407ea1f66f348636a68e0d1c6cf914c65/common/src/utils/mockPassportData.ts#L105

If you want to continue working on this, you can now try doing the full implementation for rsassaPss using your mgf1

This would be super valuable, and also a nice way to test it as it's hard for us to know if your implementation follows the same one used in passports unless we test it with a compliant passport data like the one I just added.

Issue #38, opened 2023-12-31
Add support for new signature algorithms


0xturboblitz:
It's crucial for proof of passport to support more signature algorithms. See the map [here](https://map.proofofpassport.com/) for which country uses which. Here there are in order of priority.

| Signature algorithm | Mock passport data | Register circuit | DSC circuit | Implemented in app | Assigned |
|-------------|---------|-----------------|-------------|---------------------|----------|
| rsa sha1 | ✅ | ✅ | ✅| ✅| |
| rsa sha256 | ✅ | ✅ |✅ |✅ | |
| rsa sha512 | | | | | |
| rsassaPss sha256 | ✅ | 🚧 | 🚧 | | |
| ecdsa sha1 | 🚧 | ✅| | | |
| ecdsa sha256 | 🚧 |✅ | | | |
| ecdsa sha384 | ✅ | | | | |
| ecdsa sha512 | 🚧| | | | |

They are all based on the same primitives (RSA, ECDSA and SHA) that all have implementations in circom.

With the new circuit design we shipped, adding support for new signature algorithms can be done easily by copying the [passport verifier circuit](https://github.com/zk-passport/proof-of-passport/tree/main/circuits/circuits) and modifying it.

Hash function circuits [here](https://github.com/bkomuves/hash-circuits)

If you try your hands at those, please include typescript tests for them.

puneeth072003:
I'd like to take up this issue.

remicolin:
Hi @puneeth072003 you can work on this, have you experience building circom circuits?
You can't take the whole issue but you can start by working on one of the 3 signature/hash algorithm.

remicolin:
Hi @puneeth072003 you can work on this, have you experience building circom circuits?
You can't take the whole issue but you can start by working on one of the 3 signature/hash algorithm.

RajeshRk18:
Can I take up this issue? I've experience contributing to PLUME with circom. So, I'll be able to implement all 3 signature schemes

0xturboblitz:
Hi @RajeshRk18, yes go for it
If you have questions you can dm us at @FlorentTavernier or @colinremi on telegram

remicolin:
hi @RajeshRk18 
do you have any updates to share with us?
We will assign the rsa-pss part to the person currently working on mgf1.

RajeshRk18:
I am on it! Started working on rsapss already. Will make a pr soon

RajeshRk18:
RsassaPss is in wip.

[...]

PR #121, opened 2024-06-08
Feat: sha1 with rsa encryption

seshanthS:
Implements `sha1WithRSAEncryption` in #38


0xturboblitz:
Temporarily, uncomment sha1 test leaf for tests to pass, as they make the app registration fail for now
Will fix in next contract deployment

PR #123, opened 2024-06-09
Contribute

remicolin:
This pull request enables users to securely submit their passport data by implementing AES-256-CBC encryption with dynamically generated keys and IVs.
Additionally, the AES key is encrypted with RSA-OAEP using the server's public key.


0xturboblitz:
lgtm

PR #124, opened 2024-06-10
add NextScreen and RegisterScreen

remicolin:


remicolin:
Add NextScreen and RegisterScreen to the mobile app.

PR #139, opened 2024-06-27
Prove name is not in Ofac List

ashpect:
Resolves issue [120](https://github.com/zk-passport/proof-of-passport/issues/120)

Using smt, prove name not in ofac list by 3 level matching:
- [x] passport (100% match)
- [x] name and dob combo (High probability match)
- [x] name (Partial Match)

Add Tests
- [x] Valid passport
- [x] Ofac non-membership


remicolin (approved):
lgtm

Issue #37, opened 2023-12-31
Certificate chain verification in circom


0xturboblitz:
The electronic passports trust chain has two levels: 
- top-level certificates (CSCAs) are used to sign intermediate certificates (DSCs) 
- intermediate certificates (DSCs) are used to sign passports

Right now, we get the [list of DSCs from the ICAO website](https://github.com/zk-passport/proof-of-passport/blob/main/registry) and check if passports are signed by [any of them](https://github.com/zk-passport/proof-of-passport/blob/main/registry/outputs/signature_algorithms.json). However, we strongly suspect this list is incomplete as countries have no obligation to transmit these intermediate certificates, and the top ones are supposed to be the authoritative source.

We should work on verifying the whole certificate chain, which follows the X.509 format. The MynaWallet team has a circuit in which they are [parsing one with fixed indexes](https://github.com/MynaWallet/monorepo/blob/nico/selective-disclosure/packages/circom-circuit/src/selective-disclosure.circom), but we should build a more modular chain certificate verifier that supports multiple signature algorithms and parsing any data from certificates. This would also be super useful to projects like zk AppAttest and zk DNS trust chain verification that also require X.509 chain verification.

[Some work has been done](https://github.com/zkCert/halo2-zkcert) by Richard and Sachin in halo2.

The endgame for us would be having only the root of a merkle tree containing the [CSCAs from the masterlist](https://download.pkd.icao.int) on chain, then verifying the whole chain in the circuit.

mdabir1203:
From what I understood I have two questions : 

- by root of merkle tree means its the **hash** of all the documents from csca , right ?
- Can you tell me what is the structure of the merkle tree for CSCA master list ?

remicolin:
update on the issue:
sha256 rsa ✅ -> [code](https://github.com/zk-passport/proof-of-passport/blob/dev/circuits/circuits/dsc/dsc_sha256_rsa.circom)
sha1 rsa ✅ -> [code](https://github.com/zk-passport/proof-of-passport/blob/dev/circuits/circuits/dsc/dsc_sha1_rsa.circom)
sha256 rsapss 🚧 -> [issue](https://github.com/zk-passport/proof-of-passport/pull/141)

TODO (help wanted)
sha384 ecdsa (secpr1)

zarah-s:
I am applying to this issue via [OnlyDust platform](https://app.onlydust.com/p/proof-of-passport).

### My background and how it can be leveraged
[...]

PR #143, opened 2024-06-30
DSC/CSCA verification - prover

remicolin:
Add two major features:

- DSC verification against CSCA - sha256 with RSA encryption, key length < 4096 bits
- prover running on Modal server to generate the DSC verification proof

Other implementations:

- creation of splitBytesToWords and splitSignalsToWords circuits
- update of leaf generation with new leafHasher circuit
- new scripts for registry
- update of contracts test logic
- add mock dsc/csca certificates


remicolin:
close #37 for sha256 with rsa encryption csca certificates with a key < 4096 bits

PR #144, opened 2024-07-02
Tests for common

danzipie:
A small test fix, add unit tests for common


0xturboblitz:
Thanks for the help!
Closing now. Many things changed since and we now test everything including this kind of formatting.

PR #159, opened 2024-07-28
ECDSA With SHA1 & SHA256 Support

0xVikasRushi:
# ECDSA With SHA1 & SHA256 Support

This pr implements passport verification using P256 ECDSA.

### 1. P256 ECDSA Imports
- Added necessary imports to support P256 ECDSA operations.

### 2. Input Processing and Leaf Calculation (`getLeaf` function)
- Implemented ECDSA cases.
- For ECDSA:
  - Extracts x and y coordinates from the public key.
  - Uses Poseidon6 to hash x and y coordinates separately.
  - Applies Poseidon3 to combine hashed coordinates with signature algorithm identifier.

### 3. Verifier Circuit
- Added`passport_verifier_ecdsaWithSHA1Encryption.circom`
- Added `passport_verifier_ecdsaWithSHA256Encryption.circom`
- Uses `dsc_modules[2][k]` structure where k is the number of bits.
- Processes `signature_r` and `signature_s` components.

### 4. Register Circuit 
- Added`register_ecdsaWithSHA1Encryption.circom`
- Added `register_ecdsaWithSHA256Encryption.circom`
- commitment hash similar to typescript leak calculation
- Added Verifier circuit logic for signature constraint
- Added test for valid commitment calculation, invalid signature, econtent and mrz checks


remicolin:
lgtm

PR #169, opened 2024-08-08
[fix] correcting the use of an deplicated method/libs/deps: checkout@v3 | setup-node@v3

whonion:
read more here pls: https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/)


0xturboblitz:
thanks! Closing because this has changed since

PR #209, opened 2024-10-06
fix script for build-dsc-circuit

motemotech:
It doesn't work the script to build dsc circuit in my local since the change in directory structure and file name.


motemotech:
I just merged this change into this pr.
https://github.com/zk-passport/openpassport/pull/211

PR #210, opened 2024-10-06
added simple test code for sbt contract and mock contracts

motemotech:
Added test code to check 
- validation to mint
- These functions working correctly
  - isStringEqual
  - substring
  - getCurrentTimestamp


motemotech:
I guess I should not have single test file and should merge to one single test file.

PR #213, opened 2024-10-06
Fix/test refactor prepare flow

motemotech:
This pr is on top of the change in this pr
https://github.com/zk-passport/openpassport/pull/212

- fix path to the right place of circuit output
- fix impors which is inside RegisterAndDisclose.ts
- fix function calls in preparation phase in test


motemotech:
This pr doesn't make sense that much

PR #214, opened 2024-10-07
Feat/one time sbt contract

motemotech:
- Add OneTimeSBT.sol
  - This is the contract you can get verified attr after your pass thorough these verification below
    - Prove circuit
    - DSC circuit
    - Current time verification
    - blinded dsc commitment equality
  - Also, I added expiration date for this OneTimeSBT. I just set for 90 days, since CRL is published by ICAN every 90 days.
- Add VerifiersManager.sol
  - This is the contract to manage different verifiers for each process and signature algorithm


contracts/contracts/interfaces/IVerifiersManager.sol
```diff
@@ -0,0 +1,50 @@
+//SPDX-License-Identifier: MIT
+pragma solidity ^0.8.18;
+
+interface IVerifiersManager {
+
+    error ZERO_ADDRESS();
+
+    struct RSAProveCircuitProof {
```
remicolin, line 8: change 'RSAProveCircuitProof' for 'ProveCircuitProof' as ecdsa signature is also used to sign passports

PR #216, opened 2024-10-07
fix: change signature algorithm signal output in prover circuits

AngeloAyranji:


AngeloAyranji:
I updated the signatureAlgorithm output signal in the prover circuits to use the correct number, aligning it with the values defined in constants.js located in the common directory. These values also match the ones used in the register circuits.

0xturboblitz:
thanks! Closing because this has changed since

PR #292, opened 2025-01-07
fix rsapss 4096 bits

seshanthS:


seshanthS:
Merged in https://github.com/zk-passport/openpassport/pull/301

PR #358, opened 2025-02-04
Sanity checking pubkey prefixes

0xturboblitz:
Checking pubkey prefixes for RSA.
```
2048 bits: [0x02, 0x82, 0x01, 0x01, 0x00]
3072 bits: [0x02, 0x82, 0x01, 0x81, 0x00] 
4096 bits: [0x02, 0x82, 0x02, 0x01, 0x00]
```
Not checking for ECDSA as the chances of finding a valid EC point are low.

Added scripts to identify sequences in certificates. It seems that no currently known certificate includes two valid prefixes.


0xturboblitz:
Moved to other repo

Issue #369, opened 2025-03-05
Refactor secret and passportData management


0xturboblitz:
Secret and passport data management is currently messy. There are many functions that have small variants, often have to handle JSON parsing and null/false returns.

Because this is so sensitive, right now we're trying never to change those functions to avoid breaking anything, but at one point we should refactor all of that on a branch and QA all the flows before merging.

Two useful facts:

Except in the backup/recovery flow, we always need both secret and passport data at the same time. It is thus simplest to have functions that return both.
We do not want to always require biometric authentication to access them, as they are often required for navigation. For instance, to know if the user is registered and if their passport is nullified onchain.

PR #387, opened 2025-03-07
remove biometrics.createKeys()

0xturboblitz:
I’m getting `Error generating public private keys` on open on a Redmi 14C.

It looks like we used to use `biometrics.createKeys()` and `biometrics.createSignature()`; instead of simply calling `biometrics.simplePrompt()` to check biometrics.
It’s unclear to me why we needed to create a keypair and sign the passport data/secret when it’s retrieved from the keychain, especially since it looks like the signature was always disregarded.
Rémi solved issues he was having on Android [here](https://github.com/selfxyz/self/commit/14495732a3112b62dbf49fdbfd047c1765095138) by replacing `biometrics.createSignature()` by `biometrics.simplePrompt()`
This PR also removes `biometrics.createKeys()` which causes issues on my side.

Wanted to confirm first with @nicolasbrugneaux before merging anything as I don't want to break stuff.

### DIFF
--- a/app/src/screens/SplashScreen.tsx
+++ b/app/src/screens/SplashScreen.tsx
@@ -21,10 +21,7 @@ const SplashScreen: React.FC = ({}) => {
     checkBiometricsAvailable()
       .then(setBiometricsAvailable)
       .catch(err => {
-        console.warn(
-          'Error checking biometrics availability',
-          err,
-        );
+        console.warn('Error checking biometrics availability', err);
       });
   }, []);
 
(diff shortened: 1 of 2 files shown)


0xturboblitz:
Would probably also make sense to refactor more and remove the `SignedPayload` type, c.f. also https://github.com/selfxyz/self/issues/369

nicolasbrugneaux:
I believe you can go ahead if you think the UX will be better, the idea behind it was to setup grounds for the ability to verify the passport data was accessed by the user and noone else, but it does come with UX compromises, let's go ahead!

0xturboblitz:
thanks @nicolasbrugneaux

0xturboblitz:
we should do some QA before merging, but that would fix the app for my Redmi 14C for instance.

0xturboblitz:
Integrated in https://github.com/selfxyz/self/pull/459

PR #448, opened 2025-03-14
feat: automate mobile builds

transphorm:
still need to update the boilerplate code and add the secrets to github

### DIFF
--- a/app/fastlane/Fastfile
+++ b/app/fastlane/Fastfile
@@ -146,14 +146,14 @@ platform :ios do
       output_directory: "build",
       clean: true,
       # Remove export_options related to signing, let xcargs handle it
-      # export_options: {
-      #   # signingStyle: "manual", # redundant with xcargs
-      #   teamID: ENV["IOS_TEAM_ID"], # redundant with xcargs
-      #   provisioningProfiles: { # redundant with xcargs
-      #     ENV["IOS_APP_IDENTIFIER"] => ENV["IOS_PROV_PROFILE_NAME"]
-      #   }
-      # },
-      xcargs: xcargs_array.join(" ")
+      export_options: {
+        # signingStyle: "manual", # redundant with xcargs
+        teamID: ENV["IOS_TEAM_ID"], # redundant with xcargs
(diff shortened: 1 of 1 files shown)


app/Gemfile
```diff
@@ -1,8 +1,10 @@
 source 'https://rubygems.org'
 
 # You may use http://rbenv.org/ or https://rvm.io/ to install and use this version
-ruby ">= 2.6.10"
+ruby ">= 3.2"
```
transphorm, line 4: while testing using `act`, it broke on earlier versions. `3.2` is the lowest stable version that is currently active

app/Gemfile.lock
```diff
@@ -1,27 +1,57 @@
 GEM
   remote: https://rubygems.org/
   specs:
-    CFPropertyList (3.0.6)
+    CFPropertyList (3.0.7)
```
transphorm, line 4: ran `bundle update`

app/.ruby-version
```diff
@@ -0,0 +1 @@
+3.2.7
```
transphorm, line 1: be explicit and set ruby version, set it to lowest available stable version

app/android/build.gradle
```diff
@@ -2,10 +2,10 @@
 
 buildscript {
     ext {
-        buildToolsVersion = "34.0.0"
+        buildToolsVersion = "35.0.0"
```
transphorm, line 5: build notes reported that lowest supported version is 35

app/ios/Self.xcodeproj/project.pbxproj
```diff
@@ -481,7 +481,7 @@
 				CODE_SIGN_ENTITLEMENTS = OpenPassport/OpenPassportDebug.entitlements;
 				CODE_SIGN_IDENTITY = "Apple Development";
 				CODE_SIGN_STYLE = Automatic;
-				CURRENT_PROJECT_VERSION = 105;
+				CURRENT_PROJECT_VERSION = 107;
```
transphorm: match what's currently in the app store

PR #459, opened 2025-03-14
fix: refactor secret and passport-data management

### LINKED ISSUE #369: Refactor secret and passportData management

ehsunahmadi:
Closes #369  by refactoring data handling in authentication flows.  
---  
**Changes**  
- Implemented in-memory caching for `usePassport`/`useAuth` hooks to reduce repetitive database reads  
- Added `status` state to `usePassport` (`initializing`/`updating`/`success`) for better loading state handling  
- Removed data duplication by moving mnemonic/secret logic from `useAuth` to `usePassport`  
- Revised authentication flow to prevent redundant validation checks  
- Refactored `Mnemonic.tsx` to directly consume data from `usePassport`  
- Deprecated legacy `get/create` methods in favor of memory-cached hooks  
- Some other minor bugs were resolved while tracing the data flow.

**Notes**  
- Components using auth/passport data should now rely on `usePassport`  
- Check `status` property when rendering dependent UI elements  
- Keychain interactions now only occur during initial fetch/writes


0xturboblitz:
hey @ehsunahmadi 
I did this recently, thinking maybe it could be combined with your pr?
https://github.com/selfxyz/self/pull/387

ehsunahmadi:
hey @0xturboblitz it definitely should be combined. either merge yours first and let me rebase with main or I can cherry-pick your changes in my own pr.

ehsunahmadi:
I wrapped up my work and it can be merged. @0xturboblitz let me know if I should add the changes made in https://github.com/selfxyz/self/pull/387 to this too, so we don't create signing keys and only check for biometrics.

ehsunahmadi:
ok @0xturboblitz , I removed the `createKeys` and saved availblity in useAuth hook instead of useSetting. I suggest to thin out the setting state and not copy read-only data in it as It could lead to synchronization issues down the line.
the other thing to mention about my approach to refactoring: before in these hooks, and at other places still now, there is an ineffective pattern where the hooks are returning a utility function without managing the side-effects themselves. using hooks for this purpose introduces unnecessary complexity with little to no benefit – if organization is the goal, regular module exports work just better. hooks become valuable tools for data-fetching, when we use them to manage the lifecycle of fetching the data and setting it inside the hook and create states for the values and their status through the lifecycle ("fetching", "loading", "success" etc) so they can be handled consistently everywhere.
if you check the changes in the pr, I used this approach a bunch, and replaced the direct get-functions being returned from hooks with the actual values and their status. the calling and managing of the functions was handled inside the hooks themselves. I did the same when applying the changes from your pr. the check for availblity will happen as soon as the auth-provider is rendered in the root and the state is saved for anyone to read from later on, so the useEffect was deleted entirely.

0xturboblitz:
Thanks! Gotta test it a bit more before merging but it looks pretty good

remicolin (approved):

PR #470, opened 2025-03-19
chore: fix inconsistent quote usage

zkpepe:
**Description:**  
I noticed that double quotes (`"`) were used in a few places in `script.js`, while the rest of the file consistently uses single quotes (`'`).


0xturboblitz:
lame.

PR #473, opened 2025-03-22
Fix/refactor secret and passport data management

### LINKED ISSUE #369: Refactor secret and passportData management

0xturboblitz:
Recreated https://github.com/selfxyz/self/pull/459 by @ehsunahmadi, that we merged too fast then had to revert. We will setup a staging branch to avoid this issue in the future.

One of the thing it breaks is that `secret.password` used at multiple place of the app is the password from the `Mnemonic` object in ethers (which is always "" in our case) instead of the private key of the first wallet of the HDWallet, e.g. from `const wallet = ethers.HDNodeWallet.fromPhrase(mnemonic.phrase)`.

Other quick notes:
- use `usePassport(false)` in `PassportNFCScanScreen.tsx` and `LoadingScreen.tsx`
- looks like `Stored mnemonic parsed successfully` is logged twice on open

Copying previous description here:

Closes #369  by refactoring data handling in authentication flows.  
---  
**Changes**  
- Implemented in-memory caching for `usePassport`/`useAuth` hooks to reduce repetitive database reads  
- Added `status` state to `usePassport` (`initializing`/`updating`/`success`) for better loading state handling  
- Removed data duplication by moving mnemonic/secret logic from `useAuth` to `usePassport`  
- Revised authentication flow to prevent redundant validation checks  
- Refactored `Mnemonic.tsx` to directly consume data from `usePassport`  
- Deprecated legacy `get/create` methods in favor of memory-cached hooks  
- Some other minor bugs were resolved while tracing the data flow.

**Notes**  
- Components using auth/passport data should now rely on `usePassport`  
- Check `status` property when rendering dependent UI elements  
- Keychain interactions now only occur during initial fetch/writes


ehsunahmadi:
Hey @0xturboblitz,

I’ve addressed the issues you mentioned. I replaced `secret.password` with the privateKey and used `usePassport(false)` in both `PassportNFCScanScreen.tsx` and `LoadingScreen.tsx`. I also merged the current main branch into it. Let me know if there are any other issues.

Also, I wasn't able to reproduce the problem you mentioned about the "Stored mnemonic parsed successfully" being logged twice on open. It only logs once for me when I go through the mockPassport flow.

remicolin:
hey @ehsunahmadi another small thing, last time I tested, on `ProveScreen` and `ShowRecoveryPhrase` the biometrics were triggered at the moment you land on the screen. we want to do the biometric check only when you perform the action, like clicking on the button to see the recovery phrase or send the Disclosure payload. I think we can just request the passportData without biometrics in both screen and require to pass the biometrics only when you perform the action.

PR #481, opened 2025-03-25
Fix input generation for 521bit curves

0xturboblitz:
- fix EC point padding for 521 bit curves
- rename modulus to point in findStartIndexEC as it is a point
- simplify logic


0xturboblitz:
ready for review

Nesopie (approved):

PR #484, opened 2025-03-27
Feat: Continue Register flow

seshanthS:
##changes

If the user exits the app during the registration flow, It navigates to LoadingScreen.

- Navigates directly to LoadingScreen from splashScreen
- `Get Started` button in LaunchScreen navigates to LoadingScreen if registration is in progress
- Reconnects websocket in LoadingScreen when the error code is 1006, Instead of navigating user back to LaunchScreen.


0xturboblitz:
I think it makes more sense to check for the registration status in SplashScreen with the rest of the tests than in LaunchScreen

0xturboblitz:
I see.
Ideally we would not redirect the user to the LaunchScreen directly on registration error, but to a screen giving more detail about the error, c.f. https://github.com/selfxyz/self/issues/428. This is already an improvement tho.

0xturboblitz:
In general when it's possible, we want to avoid passing side effects functions like `setRegistrationSessionId` and `setRegistrationFlowStatus` to a function like `registerPassport`, instead either use the return value of `registerPassport` or  call the function from the zustand store directly using `<store>.getState().<function>`

PR #495, opened 2025-03-30
generate scope from dns

remicolin:


common/src/utils/scope.ts
```diff
@@ -0,0 +1,63 @@
+import { poseidon2 } from "poseidon-lite";
+import { flexiblePoseidon } from "./hash";
+
+export function formatEndpoint(endpoint: string): string {
+    if (!endpoint) return '';
+    return endpoint
+        .replace(/^https?:\/\//, '')
```
motemotech, line 7: Endpoint can be contract address. 

This is the changes which I worked on before
https://github.com/selfxyz/self/blob/2b05998e6093a3e126959e97fc990aed53c573a2/common/src/utils/circuits/generateInputs.ts#L353C1-L362C2

motemotech: Sry, seems we don't need ton consider this

motemotech (approved):

PR #515, opened 2025-04-14
fix: incorrect use of `perspective` on the same element

mdqst:
`perspective` was being applied directly to the element being transformed, which doesn't have any effect according to the CSS spec.
`perspective` should be set on the parent instead.

updated the structure so that transforms now happen on a child element with the perspective correctly applied to its parent.


remicolin:
closing because of the merging issue on dev branch

PR #523, opened 2025-04-17
Aadhar Intergrations

RazorClient:
This pr involves extending self to include other government ids ,starting with aadhaar

This PR is a WIP 

there a few items that are being worked on

- [ ] Evolving Anon Aadhaars Fire and Forget to the register +disclosure self follows
-  An important blocker here is the fields used to derive the attestation nullifier.
aadhar used the photodata ,but due to their fire and forget type of proving ,it doesnt hurt that photos can be changed 
however dues to the 2 different nullifier we have ,with the majority of the problems with the attestation nullifier,the nullifier cant be inputted with photohash and the other feilds like name,dob etc dont have entropy enough to be used on their own 

- [ ] need to work on a template that does name extraction 
- [ ] other questions to be added here

### DIFF
--- a/circuits/circuits/disclose/aadhar_vc_and_disclose.circom
+++ b/circuits/circuits/disclose/aadhar_vc_and_disclose.circom
@@ -1,0 +1,119 @@
+pragma circom 2.1.9;
+
+include "circomlib/circuits/poseidon.circom";
+
+
+/// @title AADHAAR_VC_AND_DISCLOSE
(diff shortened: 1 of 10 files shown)


RazorClient:
https://www.notion.so/Indian-identity-Integration-1dc57801cd1280bebd45f3527ef60150
for context

RazorClient:
@remicolin where do i keep the file to generate the newer ofac_smts
register is no longer in the dev branch

PR #532, opened 2025-04-24
display error animation on failure on loading screen

remicolin:


app/src/utils/proving/provingMachine.ts
```diff
@@ -338,6 +341,7 @@ export const useProvingStore = create<ProvingState>((set, get) => {
         const data =
           typeof message === 'string' ? JSON.parse(message) : message;
         console.log('Received status update with status:', data.status);
+        console.log(data);
```
transphorm: @remicolin did we still want to log data here? or can we remove?

PR #536, opened 2025-04-27
Push notification

motemotech:


app/android/app/build.gradle
```diff
@@ -142,6 +143,13 @@ dependencies {
 
     implementation 'com.github.blikoon:QRCodeScanner:0.1.2'
 
+    // Firebase Dependencies
+    implementation platform('com.google.firebase:firebase-bom:32.7.3')
+    implementation 'com.google.firebase:firebase-analytics'
+    implementation 'com.google.firebase:firebase-messaging'
+    implementation 'com.google.firebase:firebase-installations:17.1.4'
+    implementation 'com.google.firebase:firebase-iid:21.1.0'
```
remicolin: are we using all these implementation?

common/src/constants/mockCertificates.ts
remicolin, line 1: let's try to avoid pushing updates on this file and discard the changes

remicolin:
image

Facing this issue when building the app on iOS

common/src/constants/mockCertificates.ts
motemotech: okay

app/android/app/build.gradle
motemotech: I just pushed to delete some implementations from build.gradle, but I can't test it rn in my environment.
Will try it tonight.

PR #553, opened 2025-05-13
Claude Code: Add missing events and tests

justinhernandez:


justinhernandez:
it was fun..don't think i'll be using claude code for writing new code. just tests and docs imo

PR #569, opened 2025-05-20
Feature/implement self verification consumer

kevinsslin:
# Add SelfVerificationConsumer for improved developer experience

## Changes

- Added `SelfVerificationConsumer` abstract contract to simplify integration with Self verification
  - Built on top of `SelfVerificationRoot` with automatic nullifier tracking
  - Provides convenient hooks for implementing custom business logic
  - Streamlines verification workflow with proper parameter passing

- Enhanced HappyBirthday example contract to showcase the new pattern
  - Simplified implementation using the consumer contract
  - Added configurable parameters (claimable amount and window)
  - Used custom errors for better gas efficiency
  - Fixed user identification to properly send rewards to the verified user

## Integration Benefits

The `SelfVerificationConsumer` offers three key integration points that simplify development:

1. **onVerificationSuccess Hook**:
   - Core integration point for developers to implement business logic
   - Automatically receives revealed data and user identifier after successful verification
   - Allows developers to focus solely on what happens after verification
   - Example: HappyBirthday simply checks birthday window and transfers tokens

2. **Customizable Nullifier Management**:
   - Default implementation provides one-time-use nullifiers (most common case)
   - `validateNullifier`: Override to implement custom validation rules
     - Allow temporary reuse (e.g., within a time period)
     - Implement usage counts (e.g., allow 3 uses per nullifier)
   - `updateNullifier`: Override to customize how nullifier state changes
     - Track usage counts or timestamps
     - Add expiration mechanics
     - Implement different types of limits

3. **Streamlined Inheritance Pattern**:
   - Developers only need to inherit from `SelfVerificationConsumer` instead of directly from `SelfVerificationRoot`
   - No need to manually track nullifiers, validate proofs, or handle verification errors
   - Most simple implementations only need to override `onVerificationSuccess`


contracts/contracts/abstract/SelfVerificationRoot.sol
```diff
@@ -69,27 +69,53 @@ abstract contract SelfVerificationRoot is ISelfVerificationRoot {
...
+     * @param _identityVerificationHubAddress The address of the Identity Verification Hub.
+     * @param _scopeValue The expected proof scope for user registration.
+     * @param _attestationIds The expected attestation identifiers required in proofs.
      */
-    constructor(address identityVerificationHub, uint256 scope, uint256[] memory attestationIds) {
-        _identityVerificationHub = IIdentityVerificationHubV1(identityVerificationHub);
-        _scope = scope;
-        for (uint256 i = 0; i < attestationIds.length; i++) {
-            _attestationIds[attestationIds[i]] = true;
+    constructor(address _identityVerificationHubAddress, uint256 _scopeValue, uint256[] memory _attestationIds) {
```
motemotech: For the arguments declared in our contract, I just followed this pattern
- Function Argument Names
https://docs.soliditylang.org/en/latest/style-guide.html#function-argument-names
- Avoiding Naming Collisions
https://docs.soliditylang.org/en/latest/style-guide.html#avoiding-naming-collisions
- Underscore Prefix for Non-external Functions and Variables
https://docs.soliditylang.org/en/latest/style-guide.html#underscore-prefix-for-non-external-functions-and-variables

It seems not all contracts in production follows this rules. Actually, the rules are so much variety, but at least, we should follow this rule.

contracts/contracts/abstract/SelfVerificationRoot.sol
```diff
@@ -104,73 +130,102 @@ abstract contract SelfVerificationRoot is ISelfVerificationRoot {
...
 
     /**
      * @notice Verifies a self-proof
      * @dev Validates scope and attestation ID before performing verification through the identity hub
-     * @param proof The proof data for verification and disclosure
+     * @param _proof The proof data for verification and disclosure
      */
-    function verifySelfProof(ISelfVerificationRoot.DiscloseCircuitProof memory proof) public virtual {
-        if (_scope != proof.pubSignals[CircuitConstants.VC_AND_DISCLOSE_SCOPE_INDEX]) {
+    function verifySelfProof(ISelfVerificationRoot.DiscloseCircuitProof calldata _proof) public {
```
motemotech: Maybe we should get bytes data for here to have forward compatibility in some id systems?

kevinsslin: yes, for sure, I think that's the next step, and should be in another PR!

contracts/contracts/abstract/SelfVerificationRoot.sol
kevinsslin: Hey @motemotech, please check [bac5e6d](https://github.com/selfxyz/self/pull/569/commits/bac5e6d5138cca930132dd83e17a33b0c7e5f6f6), I’ve reverted all the prefix changes for function arguments.

contracts/contracts/example/Airdrop.sol
```diff
@@ -89,17 +95,17 @@ contract Airdrop is SelfVerificationRoot, Ownable {
...
+     * @param _identityVerificationHubAddress The address of the Identity Verification Hub.
+     * @param _scopeValue The expected proof scope for user registration.
      * @param _attestationIds The expected attestation identifiers required in proofs.
      * @param _token The address of the ERC20 token for airdrop.
      */
     constructor(
-        address _identityVerificationHub,
-        uint256 _scope,
+        address _identityVerificationHubAddress,
+        uint256 _scopeValue,
```
motemotech: Here still "_" prefix is left, but the other part is lgtm!

motemotech:
@kevinsslin 
After you change that part then you can just merge this by yourself.

contracts/contracts/example/Airdrop.sol
kevinsslin: [09355e6](https://github.com/selfxyz/self/pull/569/commits/09355e62066941f1f5d26112778f87de2723a96a)

PR #573, opened 2025-05-26
contracts: feat/bytes inputs

motemotech:
- one verifyVcAndDisclose function with bytes data
- unified library for passport and eu id card
- unified constant.sol for passport and eu id card


contracts/contracts/constants/CircuitConstantsV2.sol
```diff
@@ -47,80 +48,59 @@ library CircuitConstantsV2 {
...
+        uint256 forbiddenCountriesListPackedIndex;
+        uint256 nullifierIndex;
+        uint256 attestationIdIndex;
+        uint256 merkleRootIndex;
+        uint256 currentDateIndex;
+        uint256 namedobSmtRootIndex;
+        uint256 nameyobSmtRootIndex;
+        uint256 scopeIndex;
+        uint256 userIdentifierIndex;
+        uint256 passportNoSmtRootIndex; // Only for passport, 0 for ID card
```
kevinsslin, line 64: What does this mean? can we have comment on this variable for readability?

contracts/contracts/constants/CircuitConstantsV2.sol
```diff
@@ -47,80 +48,59 @@ library CircuitConstantsV2 {
...
+    }
+
+    /**
+     * @notice Returns the circuit indices for a given attestation type.
+     * @param attestationId The attestation identifier.
+     * @return indices The DiscloseIndices struct containing all relevant indices.
+     */
+    function getDiscloseIndices(bytes32 attestationId) internal pure returns (DiscloseIndices memory indices) {
```
kevinsslin, line 72: What are the indices for?

PR #587, opened 2025-06-01
Feature/implement self verification root based on identity verification hub v2

kevinsslin:
# Refactor SelfVerificationRoot to Support Bytes-Based Proof Verification

## Overview
This PR introduces a major architectural upgrade to the SelfVerificationRoot contract system, migrating from struct-based to bytes-based proof verification while adding contract versioning and enhanced attestation handling.

### 🔄 **Core Architecture Updates**
- **Migration to IdentityVerificationHubV2**: Replaced `IIdentityVerificationHubV1` with `IIdentityVerificationHubV2` for improved bytes-based proof processing
- **Contract Versioning**: Added `_contractVersion` field with validation to ensure compatibility between relayer and contract versions
- **Bytes-Based Verification**: New `verifySelfProof(bytes calldata relayerData)` method that parses structured relayer data format

### 🆔 **Attestation ID System Overhaul**
- **Type Change**: Migrated attestation IDs from `uint256` to `bytes32` for better type safety and consistency
- **Enhanced Constants**: Added document-specific constants (`E_PASSPORT_REVEALED_DATA_LENGTH`, `EU_ID_CARD_REVEALED_DATA_LENGTH`)
- **Dynamic Data Handling**: Support for variable-length revealed data based on attestation type

### 📡 **Enhanced Event System**
- **Comprehensive VerificationSuccess Event**: Now includes `scope`, `attestationId`, `nullifier`, `userIdentifier`, and `revealedDataPacked[]`
- **Contract Version Events**: Added `ContractVersionUpdated` event for version tracking
- **Simplified Event Emissions**: Removed duplicate event emissions from public functions

### 🔧 **Data Format & Parsing**
- **Structured Relayer Data**: Format: `| 1 byte circuitVersion | 31 bytes buffer | 32 bytes attestationId | proof data |`
- **Hub Data Assembly**: Constructs hub-compatible data with version info and scope validation: `| 1 byte circuitVersion | 1 byte contractVersion | 30 bytes buffer | 32 bytes attestationId | 32 bytes scope | proof data |`
- **Multi-Format Support**: Handles both passport and ID card verification results with appropriate data structures

### 🎯 **Hook Function**
```solidity
function onBasicVerificationSuccess(
    bytes32 attestationId,
    uint256 scope, 
    uint256 userIdentifier,
    uint256 nullifier,
    uint256 identityCommitmentRoot,
    uint256[] memory revealedDataPacked,
    uint256[4] memory forbiddenCountriesListPacked
) internal virtual;
```

## Updated Contracts
- ✅ **SelfVerificationRoot**: Core verification logic with bytes-based processing
- ✅ **Airdrop**: Updated constructor and attestation handling
- ✅ **HappyBirthday**: Aligned with new verification structure  
- ✅ **SelfPassportERC721**: Enhanced attestation ID management

## Validation & Error Handling
- **Data Format Validation**: Ensures minimum 64-byte relayer data length
- **Contract Version Matching**: Validates version compatibility
- **Enhanced Error Types**: Added `InvalidContractVersion` and `InvalidDataFormat` errors


contracts/contracts/abstract/SelfVerificationRoot.sol
```diff
@@ -21,17 +27,21 @@ abstract contract SelfVerificationRoot is ISelfVerificationRoot {
     /// @dev Used to validate that submitted proofs match the expected scope
     uint256 internal _scope;
 
+    /// @notice The contract version for validation
+    /// @dev Used to validate the contract version in relayer data
+    uint8 internal _contractVersion;
```
motemotech, line 32: Do you think we should have _contractVersion as updatable value?
_contractVersion shows which version their contract is integrated, so if their contract is upgradeable then they also need to update _contractVersion, and immutable then this value is fixed and fine

kevinsslin: Hmm I think we should provide this possibility for them, I have `_setContractVersion` as an internal function, so if they inherit SelfVerficationRoot without implement `setContractVersion` as an external/public function, then it's immutable in that contract.

motemotech: Make sense!

PR #591, opened 2025-06-04
Bump build to test PACE polling (ios 127; android 66)

transphorm:
- rename `MIXPANEL_PROJECT_TOKEN` to `MIXPANEL_NFC_PROJECT_TOKEN`
- sort package.json files

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -76,7 +76,6 @@
     "elliptic": "^6.6.1",
     "ethers": "^6.11.0",
     "expo-modules-core": "^2.2.1",
-    "hermes-engine": "^0.11.0",
     "js-sha512": "^0.9.0",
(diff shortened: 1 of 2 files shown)


app/ios/OpenPassport/OpenPassportDebug.entitlements
```diff
@@ -29,7 +29,6 @@
 	<array>
 		<string>PACE</string>
 		<string>TAG</string>
-		<string>NDEF</string>
```
transphorm, line 32: remove `NDEF` support; in order to use it we need to raise the target sdk version from `15` to `18.2`

justinhernandez:
it looks like the circuits tests are failing on `dev` as well. going to merge since the circuit test failures aren't related to this pr

PR #614, opened 2025-06-12
NFC command chaining - IOS

seshanthS:
## Changes
-  uses command chaining in GENERAL AUTHENTICATE step of PACE
- bump ios build to `130`


transphorm:
@seshanthS can we close some of these temporary branches?

PR #630, opened 2025-06-16
Refactor UI components to new self ui package

aaronmgdr:
This begins moving components from app/ to ui/ so they they can be shared between our current mobile app and a future web based app to be viewed thru a web view embedded in a kotlin sdk. 

@selfxyz/ui will eventually hold about 90% of our react native code. 

We take advantage of our use of tamagui which has special compiler for optimizing web code from react native components.


justinhernandez:
@aaronmgdr lmk if you need help with the merge conflicts. don't mind jumping in

justinhernandez:
@coderabbitai review

PR #662, opened 2025-06-23
Demo for Native Kotlin app / Webapp view integration

aaronmgdr:


apps/SelfDemo/app/src/main/java/com/selfxyz/demo/MainActivity.kt
```diff
@@ -0,0 +1,269 @@
...
+                }
+            }
+        }
+    }
+
+    private fun setupWebView() {
+        webView.settings.apply {
+            javaScriptEnabled = true
+            domStorageEnabled = true
+            allowFileAccess = true
```
aaronmgdr, line 89: hmm we might not need this. double check

apps/SelfDemo/app/src/main/java/com/selfxyz/demo/MainActivity.kt
```diff
@@ -0,0 +1,269 @@
...
+        }
+
+        // Add JavaScript interface
+        webView.addJavascriptInterface(WebAppInterface(), "AndroidApp")
+
+        webView.webViewClient = object : WebViewClient() {
+            override fun onPageFinished(view: WebView?, url: String?) {
+                super.onPageFinished(view, url)
+                // Inject the secret securely after page loads
+                injectSecret()
```
aaronmgdr, line 101: instead of injecting. i think requesting the secret when needed will be better.

apps/SelfDemo/app/src/main/java/com/selfxyz/demo/MainActivity.kt
```diff
@@ -0,0 +1,269 @@
...
+            } else {
+                // Fallback: store in a global variable if the web app isn't ready
+                window.selfAppSecret = '$secretHash';
+            }
+        """.trimIndent()
+
+        webView.post {
+            webView.evaluateJavascript(script, null)
+        }
+    }
```
aaronmgdr, line 126: this doesnt feel nice. instead we can define a method like Android.getPrivateKey() that is callable from js side. 

out koktlin module will either A) require the caller to pass the private key in on init or B) require a function to be passed in that when called returns the private key (probably after asking for it from secure enclave)

apps/SelfDemo/app/src/main/java/com/selfxyz/demo/MainActivity.kt
```diff
@@ -0,0 +1,269 @@
...
+
+    inner class WebAppInterface {
+        @JavascriptInterface
+        fun getPrivateKey(token: String) {
+            // validate token? () and return privateKey / aka secret
```
aaronmgdr, line 131: is there any sort of authentication that we could want to do that webapp is legit version before sending privateKey to it?

PR #671, opened 2025-06-24
SEL-423 apply xcode build suggestions

transphorm:
## Summary by CodeRabbit

- **Bug Fixes**
  - The app now supports only portrait orientation on both iPhone and iPad, removing landscape mode support.

- **Chores**
  - Removed app clip embedding and related assets from the iOS project.
  - Updated platform and device targeting settings for improved build configuration.
  - Upgraded Xcode project and scheme versions for compatibility with newer Xcode releases.


justinhernandez:
fyi @remicolin removed ios landscape mode and update build settings based on xcode's recommendations

PR #679, opened 2025-06-25
Feat/android nfc retries

seshanthS:


transphorm:
@seshanthS can we close this branch?

PR #687, opened 2025-06-26
Update README: replace outdated /registry link with Self Protocol docs

GarmashAlex:
Replaced the outdated link to /registry (previously at zk-passport/openpassport) with the current Self Protocol documentation link (https://docs.self.xyz/) for accessing the parsed list of public keys. This ensures users have access to up-to-date resources.

## Summary by CodeRabbit

* **Documentation**
  * Updated the README to reference the Self Protocol documentation website for the list of countries' public keys instead of the GitHub repository.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ Not all signature algorithms are currently supported. To help us add support for
 
 #### Where can I find the countries' public keys ?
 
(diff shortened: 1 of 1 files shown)


remicolin:
the parsed list is not available in the docs.

PR #760, opened 2025-07-11
SEL-559: Update td1 regex

seshanthS:


transphorm (approved):
tested the code within the pull request and my passport scanned fine

PR #764, opened 2025-07-14
SEL-553: Show NFC Progress

seshanthS:


transphorm:
@seshanthS tested the code out and here is some feedback

- scanning is EXTREMELY fast. speed now matches iOS
- maybe we don't need to return all the feedback to the user; sending a message like "stop moving your phone" or something like should work
- after all the nfc feedback, I still needed to physically move my phone for the nfc scan to complete
- if you look at the video, the scan completes about 2 seconds. i don't physically jiggle my phone until 6 seconds.

https://github.com/user-attachments/assets/486be259-9c90-4f97-9804-bed2e9cc101e

PR #765, opened 2025-07-15
SELF-509 firebase remote config

aaronmgdr:
setup firebase remote config for web.

this could also serve as the basis for firebase messaging notifications, which i started but stashed due to difficulty with service worker and env vars. 

I also changed the ts configs for several packages to fix issue where they kept trying to find the types for firebase(which is not used by any of them)    thru implict types |

TODO 

env vars should be via `import.meta` but id like to merge #762 first since it gets import.meta working with typescript


aaronmgdr:
closing since web is not prio

Issue #770, opened 2025-07-17
NFC scan fails due to MRZ misinterpretation: "0" (zero) read as "O" (letter O)


XAbirHasan:
## Description:

When attempting to scan a valid ePassport using NFC, the app consistently fails with the following error:

```
NFC scan error
BAC failed in MUTUAL AUTH (SW = 0x6985: CONDITIONS NOT SATISFIED) (step: 2)
```

After in-depth debugging using other passport readers and low-level BAC logs, I identified that the app misinterprets the **passport number**, particularly treating the **number "0" (zero)** as the **letter "O"** during MRZ parsing.

This incorrect parsing results in **invalid session key derivation**, causing **BAC mutual authentication to fail at step 2**.

## Issue
1. Passport number like `B00055421` is **auto-filled as** `BO0055421` (with `O` instead of `0` "zero").
2. NFC scan fails at BAC step 2 with status word `0x6985`.

## Expected Behavior:

* The app should distinguish between **numeric 0 ("zero")** and **alphabetic O ("O")**.
* The MRZ line should be parsed **exactly as printed on the passport**, with check digits and characters preserved.

## Screenshots:

1. **Passport image with correct MRZ** [red mark]
<img width="600" alt="Image" src="https://github.com/user-attachments/assets/cbe43659-ccab-4743-a987-3026e919d071" />

2. **Bug in MRZ input UI** (Incorrect parsing shown) [red mark]
<img width="400" alt="Image" src="https://github.com/user-attachments/assets/ca9f3303-0e64-41ba-ac05-5d21bffb3b7f" />

remicolin:
hey @XAbirHasan can you try again from `dev` branch?

PR #784, opened 2025-07-23
SEL-576 :Fix document number

seshanthS:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved accuracy of document number extraction by ensuring only cleaned and validated document numbers are used when processing OCR results for various document types.


transphorm:
nice work @seshanthS! can we also reach out to the [issue](https://github.com/selfxyz/self/issues/770) creator [XAbirHasan](https://github.com/XAbirHasan) and confirm that the fix works?

optional: can we add any tests to ensure this functionality?

PR #811, opened 2025-07-30
SEL-541: Fix lottie views

transphorm:
## Summary by CodeRabbit

* **Chores**
  * Improved application performance by optimizing how utility modules are bundled, resulting in more efficient loading.
  * Introduced a dedicated proving utilities module to streamline internal processes and reduce bundle size.


transphorm:
Leszek is going to take over this work. Closing

PR #854, opened 2025-08-08
fix: simplify getCountryCode function return type

GarmashAlex:
Change return type from "string | string" to "string" in getCountryCode function to remove redundant type annotation.
Reopen #849 

## Summary by CodeRabbit

* **Refactor**
  * Updated the return type of the country code lookup function to consistently return a single string value.


transphorm:
@GarmashAlex thanks!

PR #855, opened 2025-08-08
feat: add liveness check

seshanthS:


transphorm:
this is not working as expected. we need to figure out how to pre-build and cache the circuit artifacts before testing

PR #866, opened 2025-08-09
fix(scripts): correct END marker parsing in addCertificatesInTs.py

Galoretka:
- Ensure END marker is searched relative to BEGIN before slicing
- Validate positions (no +len on -1) to avoid empty/incorrect slices when END is missing
- Applied to read_certificate_content and read_private_key_content

### DIFF
--- a/common/src/scripts/addCertificatesInTs.py
+++ b/common/src/scripts/addCertificatesInTs.py
@@ -1,17 +1,23 @@
 import os
 import re
 
+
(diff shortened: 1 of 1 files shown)


remicolin:
hey, could you give more context on this PR?
do you have sample of certificates that are not working with the current implementation?

transphorm:
@Nesopie can you review these changes?

Nesopie:
closing this as it's a minor refactor

PR #907, opened 2025-08-15
SDK Go version

Vishalkulkarni45:
DRAFT

## Summary by CodeRabbit

- **New Features**
  - Go SDK for verifying Self attestations with on‑chain checks and config-driven validations (age, OFAC, country exclusions), plus a BackendVerifier API.
  - Centralized 3‑letter country codes, endpoint hashing, and revealed-data unpack/format utilities.
  - Added contract ABIs and client bindings to support verification on mainnet/testnet.
  - In‑memory and pluggable config stores with action‑ID support.

- **Documentation**
  - README with setup, API overview, and examples.

- **Tests**
  - Verifier tests with a mock config store.

- **Chores**
  - Go module and example apps added.

### DIFF
--- a/sdk/sdk-go/internal/selfBackendVerifier.go
+++ b/sdk/sdk-go/internal/selfBackendVerifier.go
@@ -12,9 +12,9 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/ethclient"
 
-	self "github.com/self/sdk" // Import public API
-	commonUtils "github.com/self/sdk/common"
-	bindings "github.com/self/sdk/contracts/bindings"
+	self "github.com/selfxyz/self/sdk/sdk-go" // Import public API
+	commonUtils "github.com/selfxyz/self/sdk/sdk-go/common"
+	bindings "github.com/selfxyz/self/sdk/sdk-go/contracts/bindings"
 )
 
 // containsHexChars checks if a string contains hexadecimal characters (a-f)
--- a/sdk/sdk-go/README.md
+++ b/sdk/sdk-go/README.md
@@ -5,7 +5,7 @@ A Go SDK for integrating with the Self protocol for privacy-preserving identity 
 ## Installation
 
 ```bash
-go get github.com/self/sdk
+go get github.com/selfxyz/self/sdk/sdk-go@main
 ```
 
 ## Quick Start
@@ -20,14 +20,15 @@ import (
     "fmt"
     "log"
 
-    "github.com/self/sdk"
+    "github.com/selfxyz/self/sdk/sdk-go"
+    "github.com/selfxyz/self/sdk/sdk-go/common"
 )
 
 func main() {
     // Create a verification configuration
     config := self.VerificationConfig{
         MinimumAge:        &[]int{18}[0],                    // Require 18+ years
-        ExcludedCountries: []self.Country3LetterCode{"USA"}, // Exclude USA
+        ExcludedCountries: []common.Country3LetterCode{common.USA}, // Exclude USA
         Ofac:              &[]bool{false}[0],                // Allow OFAC flagged individuals
     }
 
--- a/sdk/sdk-go/examples/basic/go.mod
+++ b/sdk/sdk-go/examples/basic/go.mod
@@ -2,7 +2,7 @@ module basic-example
 
 go 1.23.0
 
-require github.com/self/sdk v0.0.0
+require github.com/selfxyz/self/sdk/sdk-go v0.0.0
 
 require (
 	github.com/Microsoft/go-winio v0.6.2 // indirect
@@ -31,4 +31,4 @@ require (
 	golang.org/x/sys v0.31.0 // indirect
 )
 
-replace github.com/self/sdk => ../..
+replace github.com/selfxyz/self/sdk/sdk-go => ../..
--- a/sdk/sdk-go/examples/basic/main.go
+++ b/sdk/sdk-go/examples/basic/main.go
@@ -5,8 +5,8 @@ import (
 	"fmt"
 	"log"
 
-	self "github.com/self/sdk"
-	"github.com/self/sdk/common"
+	self "github.com/selfxyz/self/sdk/sdk-go"
+	common "github.com/selfxyz/self/sdk/sdk-go/common"
 )
 
 func main() {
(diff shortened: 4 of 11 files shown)


sdk/sdk-go/common/customHash.go
```diff
@@ -0,0 +1,122 @@
...
+
+	"github.com/iden3/go-iden3-crypto/poseidon"
+)
+
+func FlexiblePoseidon(inputs []*big.Int) (*big.Int, error) {
+	if len(inputs) == 0 {
+		return nil, fmt.Errorf("no inputs provided")
+	}
+
+	if len(inputs) > 16 {
```
Nesopie: shouldn't this be > 17?

sdk/sdk-go/internal/utils/constants.go
```diff
@@ -0,0 +1,134 @@
+package utils
+
+import (
+	"self-sdk-go/internal/types"
+)
+
+const (
+	AttestationId1 types.AttestationId = 1
```
Nesopie: can you make this passport and the other one IdCard?

sdk/sdk-go/common/customHash.go
Vishalkulkarni45: we decided not to change the circuits right? so kept it same like the circuit algo

sdk/sdk-go/internal/utils/constants.go
Vishalkulkarni45: I can, but it changes the mapping a lot. I think this way is faster than using a string-to-struct mapping ?

sdk/sdk-go/internal/selfBackendVerifier.go
```diff
@@ -0,0 +1,598 @@
...
+		message = append(message, fmt.Sprintf("[%s]: %s", issue.Type, issue.Message))
+	}
+	return strings.Join(message, "\n")
+}
+
+func NewConfigMismatchError(issue []ConfigIssue) *ConfigMismatchError {
+	return &ConfigMismatchError{Issues: issue}
+}
+
+type ConfigStore interface {
```
Nesopie: move config store to a different file and and the DefaultConfigStore and InMemoryConfigStore implementations as well

sdk/sdk-go/internal/types/types.go
```diff
@@ -0,0 +1,67 @@
...
+	"math/big"
+	"self-sdk-go/common"
+)
+
+// AttestationId represents the type for attestation identifiers
+type AttestationId int
+
+// VcAndDiscloseProof represents the zero-knowledge proof structure
+type VcAndDiscloseProof struct {
+	A [2]*big.Int    `json:"a"`
```
Nesopie: accept these in string

sdk/sdk-go/internal/selfBackendVerifier.go
```diff
@@ -0,0 +1,598 @@
...
+		userIdentifierType:              userIdentifierType,
+	}, nil
+}
+
+// Verify performs the verification of attestation with the given proof and signals
+func (s *SelfBackendVerifier) Verify(
+	ctx context.Context,
+	attestationId types.AttestationId,
+	proof types.VcAndDiscloseProof,
+	pubSignals []*big.Int,
```
Nesopie: accept this in string too. When you're making an API you'll get the inputs in string. Convert them to bigint internally

sdk/sdk-go/internal/selfBackendVerifier.go
```diff
@@ -0,0 +1,598 @@
...
+		DiscloseOutput:         genericDiscloseOutput,
+		UserData: types.UserData{
+			UserIdentifier:  userIdentifier,
+			UserDefinedData: userDefinedData,
+		},
+	}, nil
+}
+
+// castToUserIdentifier converts a big integer to user identifier string based on the specified type
+func (s *SelfBackendVerifier) castToUserIdentifier(bigInt *big.Int, userIdType types.UserIDType) string {
```
Nesopie: put this in utils.go

sdk/sdk-go/internal/selfBackendVerifier.go
```diff
@@ -0,0 +1,598 @@
...
+				}
+			}
+		}
+	}
+	isMinimumAgeValid := true
+	if configErr == nil && verificationConfig.MinimumAge != nil {
+		configMinAge := *verificationConfig.MinimumAge
+		circuitMinAge := genericDiscloseOutput.MinimumAge
+		circuitMinAgeInt := 0
+		if circuitMinAge != "00" {
```
Nesopie: age is not valid if circuitMinAge == "00"

sdk/sdk-go/internal/selfBackendVerifier.go
```diff
@@ -0,0 +1,598 @@
...
+
+				var err error
+				genericDiscloseOutput, err = utils.FormatRevealedDataPacked(attestationId, publicSignals)
+				if err != nil {
+					return nil, fmt.Errorf("error formatting revealed data: %v", err)
+				}
+			}
+		}
+	}
+	isMinimumAgeValid := true
```
Nesopie, line 345: there's 2 different validations that are happening with respect to verification config. Just use one (either this one or the s.validateWithConfig())

sdk/sdk-go/internal/selfBackendVerifier.go
Vishalkulkarni45: didn't get u? ts version does the same right?

sdk/sdk-go/internal/selfBackendVerifier.go
Vishalkulkarni45: i removed this check and using only used in s.validateWithConfig()

sdk/sdk-go/common/customHash.go
Nesopie: no if you look at the typescript code it considers the switch case of 16. 

see: https://github.com/selfxyz/self/blob/cb2ef9137e64d07e98e9200490ac5052270d7fae/common/src/utils/hash.ts#L27

sdk/sdk-go/internal/utils/constants.go
Nesopie: It's better for readability. I just meant renaming the constants to something like this: 

```
const (
	Passport types.AttestationId = 1
...
```

sdk/sdk-go/internal/selfBackendVerifier.go
Nesopie: Ignore. It's handled in s.validateWithConfig()

sdk/sdk-go/common/customHash.go
Vishalkulkarni45: tru tru , will change it

PR #920, opened 2025-08-19
SDK Go version

Vishalkulkarni45:
Go sdk similar to typescript


sdk/sdk-go/internal/selfBackendVerifier.go
```diff
@@ -0,0 +1,554 @@
...
+		provider:                        provider,
+		allowedIDs:                      allowedIds,
+		userIdentifierType:              userIdentifierType,
+	}, nil
+}
+
+// Verify performs the verification of attestation with the given proof and signals
+func (s *SelfBackendVerifier) Verify(
+	ctx context.Context,
+	attestationIdStr string,
```
Nesopie: convert this to an integer

sdk/sdk-go/internal/selfBackendVerifier.go
```diff
@@ -0,0 +1,554 @@
...
+	if len(issues) > 0 {
+		return nil, self.NewConfigMismatchError(issues)
+	}
+
+	isProofValid := false
+
+	// Use the pre-calculated attestationIdBytes32 from above
+	verifierAddress, err := s.identityVerificationHubContract.DiscloseVerifier(nil, attestationIdBytes32)
+	if err != nil || verifierAddress == (common.Address{}) {
+		isProofValid = false
```
Nesopie: return an error here saying verifier not found, instead of making isProofValid = false

sdk/sdk-go/internal/selfBackendVerifier.go
```diff
@@ -0,0 +1,554 @@
...
+			isValid, err := verifierContract.VerifyProof(nil, aFormatted, bFormatted, cFormatted, publicSignalsArray)
+			if err != nil {
+				isProofValid = false
+			} else {
+				isProofValid = isValid
+			}
+		}
+	}
+
+	if forbiddenCountriesList == nil || len(genericDiscloseOutput.Nullifier) == 0 {
```
Nesopie: why check the length of the nullifier here?

sdk/sdk-go/internal/selfBackendVerifier.go
```diff
@@ -0,0 +1,554 @@
...
+		if len(userContextData) >= 128 {
```
Nesopie: why are you checking if userContextData is >= 128? It's always the case so you don't have to check it

sdk/sdk-go/internal/selfBackendVerifier.go
```diff
@@ -0,0 +1,554 @@
...
+	}
+
+	if !isForbiddenCountryListValid {
+		*issues = append(*issues, self.ConfigIssue{
+			Type: self.InvalidForbiddenCountriesList,
+			Message: fmt.Sprintf("Forbidden countries list in config does not match with the one in the circuit\nCircuit: %s\nConfig: %v",
+				strings.Join(forbiddenCountriesList, ", "), verificationConfig.ExcludedCountries),
+		})
+	}
+	genericDiscloseOutput, err := self.FormatRevealedDataPacked(attestationId, publicSignals)
```
Nesopie: you're using self.FormatRevealedDataPacked again in the verify function. Instead call it in the verify function and then pass it to this function

sdk/sdk-go/internal/selfBackendVerifier.go
```diff
@@ -0,0 +1,554 @@
...
+				genericDiscloseOutput, err = self.FormatRevealedDataPacked(attestationId, publicSignals)
+				if err != nil {
+					return nil, fmt.Errorf("error formatting revealed data: %v", err)
+				}
+			}
+		}
+	}
+
+	// Determine minimum age validity from validation issues
+	isMinimumAgeValid := true
```
Nesopie: thsi is wrong. Minimum age is valid if circuitMinAge >= configMinAge. This if statement is never run as you're returning above if len(issues) > 0

sdk/sdk-go/internal/selfBackendVerifier.go
Vishalkulkarni45: i will remove it, coderabbit wanted this to be safer

sdk/sdk-go/internal/selfBackendVerifier.go
Vishalkulkarni45: remove this block of code and hardcoded true in return

PR #924, opened 2025-08-19
Disable automatic coderabbit reviews

transphorm:


transphorm:
@aaronmgdr @remicolin in order to disable automatic pr comments you need to disable auto review which also disables auto summary.

tl;dr: pr comments are an all or nothing. you can't disable on it's own

PR #972, opened 2025-08-28
fix(noir): correct Merkle depth selection and tests for binary_merkle_root

VolodymyrBg:
- Align binary_merkle_root with SMT usage: compute node after exactly depth levels.
- Remove off-by-one and N==depth accumulation; return final computed node.
- Test helper now pads odd levels by duplicating the last leaf and returns actual depth.
- Tests updated to use real depth and deterministic leaf counts, covering N>depth.
- No API changes; behavior now matches smt_verify and TS proof generation assumptions.

### DIFF
--- a/noir/crates/dg1/src/utils/other/binary_merkle_tree.nr
+++ b/noir/crates/dg1/src/utils/other/binary_merkle_tree.nr
@@ -4,29 +4,26 @@ pub fn binary_merkle_root<let N: u32>(
     indices: [u1; N],
     siblings: [Field; N],
 ) -> Field {
-    let mut nodes: [Field; N + 1] = [0; N + 1];
-    nodes[0] = leaf;
-
-    let mut root = 0;
+    // Compute the node after applying exactly `depth` hashing steps.
+    // The loop runs over N with a selector to avoid data-dependent control flow.
+    let mut current = leaf;
 
     for i in 0..N {
-        let isDepth = if i == depth { 1 } else { 0 };
-
-        root += isDepth * nodes[i];
+        let is_active = if i < depth { 1 } else { 0 };
 
         let child_nodes = if indices[i] == 0 {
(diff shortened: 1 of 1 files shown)


noir/crates/dg1/src/utils/other/binary_merkle_tree.nr
```diff
@@ -4,29 +4,26 @@ pub fn binary_merkle_root<let N: u32>(
...
-    let mut root = 0;
+    // Compute the node after applying exactly `depth` hashing steps.
+    // The loop runs over N with a selector to avoid data-dependent control flow.
+    let mut current = leaf;
 
     for i in 0..N {
-        let isDepth = if i == depth { 1 } else { 0 };
-
-        root += isDepth * nodes[i];
+        let is_active = if i < depth { 1 } else { 0 };
```
Vishalkulkarni45: can u use if branch {} rather than using is_active

```
for i in 0..N {
   if i <depth {
      ...
     }
}
```

noir/crates/dg1/src/utils/other/binary_merkle_tree.nr
```diff
@@ -4,29 +4,26 @@ pub fn binary_merkle_root<let N: u32>(
...
 
-        nodes[i + 1] = std::hash::poseidon::bn254::hash_2(child_nodes);
-    }
+        let next_node = std::hash::poseidon::bn254::hash_2(child_nodes);
 
-    let isDepth = if N == depth { 1 } else { 0 };
-    root += isDepth * nodes[N];
+        // If this level is active (i < depth), advance; otherwise keep current.
+        current = is_active * next_node + (1 - is_active) * current;
+    }
```
Vishalkulkarni45: by using if branch u can replace these two lines with
```
current = std::hash::poseidon::bn254::hash_2(child_nodes);
```

PR #979, opened 2025-08-28
Chore fix ios nfc scanning and compiling

transphorm:


app/ios/SelfAnalytics.swift
```diff
@@ -62,4 +62,8 @@ public class SelfAnalytics: Analytics {
             ])
         }
     }
-} 
+
+    public func flush() {
```
transphorm, line 66: fyi @seshanthS added a flush method to the analytics package

PR #981, opened 2025-08-28
chore: update to macos latest large runner

transphorm:
## Summary by CodeRabbit

* **Chores**
  * CI and deployment workflows now run build-deps and the iOS build on larger macOS runners (macos-latest-large) to improve reliability and throughput; test jobs unchanged.

* **Documentation**
  * Updated CI/CD docs to reflect the move to larger macOS runners for the iOS build job.

### DIFF
--- a/.github/workflows/mobile-ci.yml
+++ b/.github/workflows/mobile-ci.yml
@@ -31,7 +31,7 @@ concurrency:
 
 jobs:
   build-deps:
-    runs-on: macos-latest
+    runs-on: macos-latest-large
     steps:
       - uses: actions/checkout@v4
       - name: Read and sanitize Node.js version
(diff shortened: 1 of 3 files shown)


.github/workflows/mobile-ci.yml
```diff
@@ -159,7 +159,7 @@ jobs:
         run: yarn test
```
transphorm, line 159: @remicolin @aaronmgdr made a small tweak to the workflows to bump the mac runner. we can't bump the mobile ci test runner yet because there is a race condition in the build step [[workflow error](https://github.com/selfxyz/self/actions/runs/17308250354/job/49146191379?pr=981)]

@seshanthS has a fix in the works for the race condition. we should bump the runner after the fix gets merged

PR #986, opened 2025-08-29
feat: increase sha256 byte size and add new rsa circuits

Nesopie:
## Summary by CodeRabbit

- New Features
  - Added support for new RSA (PKCS#1 v1.5) and RSA-PSS algorithms with 4096-bit keys, including SHA-1 and SHA-256 variants.
  - Introduced new circuit instances for DSC, Register, and Register ID using these algorithms.
  - Extended signature verification to route new algorithm IDs.

- Changes
  - Increased SHA-256 padded signed attributes length from 128 to 256 bytes.

- Tests
  - Added comprehensive tests and cases for the new RSA/RSA-PSS variants.

- Chores
  - Expanded mock certificates/keys and SKI mappings.
  - Updated certificate generation scripts.


remicolin:
lgtm

PR #1053, opened 2025-09-11
chore(mobile-sdk-alpha): remove unused tslib dependency

aaronmgdr:
* remove tslib -- seems unused

* remove react and rn from root package.json -- looks like mistake

## Summary by CodeRabbit

* **Chores**
  * Removed a runtime dependency from the mobile SDK alpha to slim dependency footprint.
  * Added a TypeScript declaration filename to git ignore so such declaration files are no longer tracked.
  * Removed root-level runtime declarations for React and React Native; their versions remain controlled via resolution settings.
  * No functional or public API changes; integrations remain compatible.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ output/*
 mobile-sdk-alpha-ci.tgz
 **/mobile-sdk-alpha-*.tgz
 /tmp/mobile-sdk-alpha*.tgz
+dataInput.d.ts


.gitignore
```diff
@@ -15,3 +15,4 @@ output/*
 mobile-sdk-alpha-ci.tgz
 **/mobile-sdk-alpha-*.tgz
 /tmp/mobile-sdk-alpha*.tgz
+dataInput.d.ts
```
aaronmgdr, line 18: looks like this is just a generated at build time file.

PR #1098, opened 2025-09-19
Allow cloning via https

aaronmgdr:
my system is set to authenticate via https. so i added a check for that in when cloning. 

## Summary by CodeRabbit

- Bug Fixes
  - Improved reliability of iOS setup by ensuring required files are writable during post-install operations, reducing edit-time errors.

- Chores
  - Enhanced private module setup to support HTTPS-based Git authentication for local development when available, with seamless fallback to existing methods.
  - Refined cloning logic to better handle CI vs. local environments with clearer logging and resilient error handling.

- Build
  - Adjusted dependency retrieval to prefer secure HTTPS access when authenticated, improving developer experience without affecting end users.

### DIFF
--- a/app/ios/Podfile
+++ b/app/ios/Podfile
@@ -26,6 +26,18 @@ if linkage != nil
   use_frameworks! :linkage => linkage.to_sym
 end
 
+def using_https_git_auth?
+    begin
+      # backticks run command in shell and capture stdout, 2>&1 captures stderr as well
+      auth_data = `gh auth status 2>&1`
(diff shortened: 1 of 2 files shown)


app/ios/Podfile
```diff
@@ -26,6 +26,18 @@ if linkage != nil
   use_frameworks! :linkage => linkage.to_sym
 end
 
+def using_https_git_auth?
```
transphorm, line 29: nice thanks for adding. we could also add support to build gradle

PR #1187, opened 2025-10-02
separate MRZ data from userStore

shazarre:


app/src/stores/userStore.ts
```diff
@@ -33,11 +26,6 @@ interface UserState {
 }
 
 const useUserStore = create<UserState>((set, _get) => ({
-  passportNumber: DEFAULT_PNUMBER ?? '',
```
shazarre, line 36: @transphorm are those `DEFAULT_*` consts something that need to be moved to the SDK or empty strings as default is fine?

app/src/utils/nfcScanner.ts
```diff
@@ -52,8 +52,6 @@ export const parseScanResponse = (response: unknown) => {
 };
 
 export const scan = async (inputs: Inputs) => {
-  await configureNfcAnalytics();
```
shazarre, line 55: this [is now called](https://github.com/selfxyz/self/pull/1187/files#diff-9fa027bd6430742b3931e0e2e7f0c77eedbc73bf0a7a48fc6466f5412532b1edR327) before calling `scan` as preparation to move the code over to the SDK

packages/mobile-sdk-alpha/src/stores/mrzStore.tsx
```diff
@@ -0,0 +1,60 @@
...
+    dateOfBirth: string;
+    dateOfExpiry: string;
+    countryCode: string;
+    documentType: string;
+  }) => void;
+  clearMRZ: () => void;
+  update: (patch: Partial<MRZState>) => void;
+}
+
+// TODO: what about the defaults from @env?
```
aaronmgdr, line 27: i think better to leave out now. that we are a production app

packages/mobile-sdk-alpha/src/stores/mrzStore.tsx
```diff
@@ -0,0 +1,60 @@
...
+  passportNumber: '',
+  dateOfBirth: '',
+  dateOfExpiry: '',
+  countryCode: '',
+  documentType: '',
+};
+
+/*
+  Never export outside of the mobile sdk. It can cause multiple instances of the store to be created.
+  Use the functions above to interact with the store.
```
aaronmgdr:
```suggestion
 interact with the store thru the self client
```

app/src/stores/userStore.ts
transphorm: yeah I think empty strings should be fine. seems like a hangnail from old logic

packages/mobile-sdk-alpha/src/stores/mrzStore.tsx
aaronmgdr: that is setting defaults from env var. that seems like something that can be removed. but @transphorm thoughts?

PR #1200, opened 2025-10-03
move common aadhaar functions to msdk

aaronmgdr:
## also
* add info that functions are duplicated and to use the msdk one 
* use listeners to navigate
* improve/fix the Sdk event descriptions


aaronmgdr:
@coderrabbit review

PR #1280, opened 2025-10-13
Move Layout Components to SDK

aaronmgdr:
refactoring with claude to allow us to shop msdk without tamagui but have a similar style. 

 Layout Components Created

  - View - Base component with full tamagui-compatible API including pressable functionality
  - XStack - Extends View with flexDirection: 'row'
  - YStack - Extends View with flexDirection: 'column'
  - Button - Extends View with button defaults (DRY implementation!)
  - Text - Standalone text component with tamagui-compatible spacing props

  Components Refactored

  - ✅ IDPickerScreen - Now uses mobile-sdk-alpha layout components
  - ✅ CountryPickerScreen - Now uses mobile-sdk-alpha layout components + ActivityIndicator
  - ✅ BaseNavBar - Uses new Button, View, XStack components
  - ✅ IdDetailsNavBar - Uses new Button, Text, View components
  - ✅ AadhaarNavBar - Uses new Button, XStack, YStack components

  Key Improvements

  - DRY Design: Button extends View instead of duplicating all layout logic
  - Full API Compatibility: All components support tamagui props like $4 spacing tokens
  - Proper Pressable: Button is always pressable with hitSlop support
  - Type Safety: Proper TypeScript with exported ViewProps
  - Token Conversion: Handles $ spacing and radius tokens correctly
  
  
  ### preview
  
  

https://github.com/user-attachments/assets/6fbfdd16-67be-446e-b84d-7a4dff9a2d81

## Summary by CodeRabbit

* **New Features**
  * Added core UI primitives for consistent UI building: View, XStack, YStack, Button, and Text.

* **Refactor**
  * Nav bars and document-selection screens now use the shared UI primitives for a unified look.
  * NavBar layout props made configurable for improved alignment and flexibility.

* **Style**
  * Replaced loading spinner with native activity indicator.
  * Tweaked document-type press animation and standardized back-button sizing.

### DIFF
--- a/packages/mobile-sdk-alpha/src/components/layout/View.tsx
+++ b/packages/mobile-sdk-alpha/src/components/layout/View.tsx
@@ -142,16 +142,16 @@ const convertBorderRadius = (
   if (value === undefined) return undefined;
   if (typeof value === 'number') return value;
 
-  if (value instanceof Animated.AnimatedNode) {
-    return value;
-  }
-
   // Handle tamagui radius tokens
   if (typeof value === 'string') {
     if (value.startsWith('$') && radiusTokens[value] !== undefined) {
       return radiusTokens[value];
     }
     return parseFloat(value) || 0;
+  }
+
+  if (typeof value === 'object' && value instanceof Animated.AnimatedNode) {
+    return value;
   }
 
   return 0;
@@ -278,7 +278,7 @@ export const View: React.FC<ViewProps> = ({
     if (Object.keys(props).length > 0) {
       console.debug('other props not handled:', props);
     }
-  }, []);
+  }, [Object.keys(props).length]);
 
   if (onPress) {
     // Convert numeric hitSlop to proper format


packages/mobile-sdk-alpha/src/components/layout/View.tsx
```diff
@@ -0,0 +1,204 @@
+// SPDX-FileCopyrightText: 2025 Social Connect Labs, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+// NOTE: Converts to Apache-2.0 on 2029-06-11 per LICENSE.
+
+import React from 'react';
+import { View as RNView, Pressable } from 'react-native';
+import type { ViewProps as RNViewProps, ViewStyle, PressableProps, DimensionValue } from 'react-native';
+
+interface SpacingProps {
```
shazarre, line 18: there's no interface we could inherit the props from (applies to other interfaces here as well)?

app/src/components/NavBar/AadhaarNavBar.tsx
```diff
@@ -4,10 +4,10 @@
 
 import React from 'react';
 import { useSafeAreaInsets } from 'react-native-safe-area-context';
-import { Button, XStack, YStack } from 'tamagui';
```
shazarre, line 7: as a generic comment: how about leaving the app as-is and applying those only to the SDK? thinking about the web version that is built via tamagui

packages/mobile-sdk-alpha/src/components/index.ts
```diff
@@ -9,6 +9,16 @@ export { default as Additional } from './typography/Additional';
...
 
+// Layout components
+export { Button } from './layout/Button';
+export { Text } from './layout/Text';
+export { View } from './layout/View';
+export { XStack } from './layout/XStack';
+export { YStack } from './layout/YStack';
+
+// Export types
+export type { ViewProps } from './layout/View';
```
shazarre: if we use them only for exported screens, there won't be a need to export them

aaronmgdr: but its needed for now

app/src/components/NavBar/AadhaarNavBar.tsx
aaronmgdr: but the point of these modifications is so that these components are ready to move to the sdk so that we can have the full screens in the sdk.

packages/mobile-sdk-alpha/src/components/index.ts
shazarre: absolutely

PR #1353, opened 2025-11-06
SELF-1139: update getUserAddress()

seshanthS:


app/src/utils/points.ts
```diff
@@ -137,6 +139,10 @@ export const getNextSundayNoonUTC = (): Date => {
   return nextSunday;
 };
 
+export const getPointsAddress = async (): Promise<string> => {
+  return getOrGeneratePointsAddress();
```
transphorm, line 143: nit: ipersonally like the double wrapping of this method but i understand that it gives us a nice exportable method. lgtm

transphorm (approved):

PR #1365, opened 2025-11-07
SELF-1150: remove deeplinkCallback from pointsSelfApp.

seshanthS:


transphorm:
thanks for the quick work @seshanthS

PR #1402, opened 2025-11-12
SELF-1177 - Separate button states during backup

seshanthS:
## Summary by CodeRabbit

* **Improvements**
  * Account recovery now tracks Turnkey and Cloud restores independently; each path shows its own progress and updates labels/buttons per method.
  * Manual recovery no longer depends on a shared restoring state, avoiding unnecessary disabling and improving availability.

* **Telemetry**
  * Added a telemetry event to record Turnkey restore failures for better monitoring.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


app/src/screens/account/recovery/AccountRecoveryChoiceScreen.tsx
```diff
@@ -236,7 +251,6 @@ const AccountRecoveryChoiceScreen: React.FC = () => {
             <SecondaryButton
               trackEvent={BackupEvents.MANUAL_RECOVERY_SELECTED}
               onPress={handleManualRecoveryPress}
-              disabled={restoring}
```
aaronmgdr, line 273: shouldnt there be something that disables the button at some point?

seshanthS: Other buttons do some action in this screen, Manual restore button simply navigates to the other screen.
We could add `!biometricsAvailable`, but that's not a blocker. It would still work if biometrics are not available.

Another option is to disable it when any one of the backup's restoring process is in progress. That would be the same as the previous behaviour without this PR.

PR #1416, opened 2025-11-13
SELF-1226 Fix: patches

seshanthS:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where non-Turnkey keys could be unintentionally deleted during key management operations. The system now properly filters and validates keys before attempting deletion, preventing accidental loss of important key data.

### DIFF
--- a/scripts/run-patch-package.cjs
+++ b/scripts/run-patch-package.cjs
@@ -90,30 +90,6 @@ if (!isExecutableAvailableOnPath('patch-package')) {
 
 // Run patch-package with better error handling
 try {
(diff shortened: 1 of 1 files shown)


scripts/run-patch-package.cjs
```diff
@@ -90,30 +90,6 @@ if (!isExecutableAvailableOnPath('patch-package')) {
 
 // Run patch-package with better error handling
 try {
-  const patchRun = spawnSync('patch-package', {
```
transphorm, line 93: thanks for removing

PR #1493, opened 2025-12-12
Add campaign card to Points component for New Year's Points Booster

remicolin:
- Integrated a new Pressable component that links to the X campaign URL.
- Added styles for the campaign card, including layout and text formatting.
- Updated imports to include Linking for URL handling.
- Placeholder URL for the campaign to be replaced with the actual link in the future.


remicolin:
cancelled

PR #1619, opened 2026-01-16
Hide unregistered docs from DocumentSelectorForProvingScreen

seshanthS:
### Description

- Hides unregistered docs from the selector in DocumentSelectorForProvingScreen
- Update `shouldSkip` condition to include registration check.

### Tested

Manually tested the following cases,
- Have a total of one document which has failed registration - should show no registered documents
- Have a total of two failed documents - Should show no registered documents
- Have a total of 2 docs with 1 not registered - DocumentSelectorForProvingScreen should show only registered document

### How to QA

1. Delete all documents
2. Register a mock document, exit when it asks for fingerprint/faceId
3. Open a proof request through deeplink
4. App should show "no documents found" instead of "Are you new here" screen
5. Register another mock document completely
6. Open a proof request through deeplink
7. The document selector should show only the selected document.

## Summary by CodeRabbit

* **Bug Fixes**
  * Document selection interface now displays only registered documents instead of all available documents
  * Auto-progression to the proving step now requires selected documents to have registered status

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/app/src/screens/verification/DocumentSelectorForProvingScreen.tsx
+++ b/app/src/screens/verification/DocumentSelectorForProvingScreen.tsx
@@ -219,6 +219,7 @@ const DocumentSelectorForProvingScreen: React.FC = () => {
 
   const documents = useMemo(() => {
     return documentCatalog.documents
+      .filter(metadata => metadata.isRegistered)
       .map(metadata => {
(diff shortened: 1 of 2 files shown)


transphorm:
nice. let's get this in to the next build

PR #1685, opened 2026-02-03
chore: self UUID namespace as a const

shazarre:
### Description

This PR removes Self UUID namespace env var in favor of having it as a const.

### Tested

Ran tests.

## Summary by CodeRabbit

* **Refactor**
  * Consolidated UUID namespace value management by centralizing the constant definition within the notifications service, replacing function-based access patterns with direct constant references throughout the application.

* **Tests**
  * Updated test mocks to rely on actual module exports instead of hand-written mock definitions.


Nesopie (approved):
lgtm

PR #1905, opened 2026-04-02
Fix recovery rollback handling

transphorm:
## Summary

- Add rollback logic to `finalizeRecoveredDocumentRegistration` so that if marking the document as registered fails after a CSCA re-store, the original document and registration state are restored
- Make `restoreSecretFromMnemonic` and `restoreStoredSecretSnapshot` capture a pre-write snapshot and attempt best-effort rollback on storage write failures
- Guard all post-async `navigate()` calls in `SecretPhraseInputScreen` with `isMountedRef` to prevent navigation after unmount
- Add `document_unavailable` error reason and explicit null-check for `loadSelectedDocument` so a missing document routes to the failure screen instead of silently proceeding
- Wrap `loadSelectedDocument` in `TourScreen` with try/catch to fall through to KYC when document state is unavailable; fix `onResore` → `onRestore` typo

### SDK Core (`mobile-sdk-alpha`)

- `recoveryValidation.ts`: wrap `finalizeRecoveredDocumentRegistration` in try/catch that rolls back document data (via `storePassportData`) and registration flag (via `updateDocumentRegistrationState`) on failure
- `recoveryValidation.test.ts`: add tests for rollback on registration failure, rollback without CSCA, and continued flag-clear when document restore fails

### WebView App

- `SecretPhraseInputScreen.tsx`: add `document_unavailable` error reason; null-check `loadSelectedDocument` result; guard all `navigate()` calls with `isMountedRef.current`
- `TourScreen.tsx`: wrap `loadSelectedDocument` in try/catch; fix `onResore` → `onRestore` typo
- `secretManager.ts`: extract `readStoredSecretSnapshotUnlocked`; add `restoreSnapshotBestEffort` for per-field rollback; make `restoreStoredSecretSnapshot` capture previous state and rollback on failure; acquire lock in `readStoredSecretSnapshot`
- `secretPhraseInputScreen.test.tsx`: new test file covering null document, mock document, unmount-during-async, and storage-write-failure scenarios
- `secretManager.test.ts`: add tests for `readStoredSecretSnapshot` lock behavior, `restoreStoredSecretSnapshot` rollback, and partial rollback when mnemonic restore fails

### Config

- `package.json`: add `test` and `test:watch` scripts to webview-app

## Test Plan

- [ ] `cd packages/mobile-sdk-alpha && yarn test` passes
- [ ] `cd packages/webview-app && yarn test` passes
- [ ] `yarn lint && yarn types` passes
- [ ] Manual: trigger recovery with invalid secret → failure screen shown, document state unchanged
- [ ] Manual: trigger recovery with valid secret → success screen, document marked registered

## Summary by CodeRabbit

* **Bug Fixes**
  * Safer recovery and restore flows: preserve prior document/registration state on failures, perform conditional rollback with best-effort restores, log rollback errors, and prevent post-unmount navigation during recovery.
  * More robust tour/navigation paths when document loading fails.

* **Tests**
  * Expanded tests covering recovery finalization, secret restore/rollback, navigation/unmount behavior, and failure/rollback scenarios.

* **Chores**
  * Added test scripts for the webview app package.


Tranquil-Flow (commented):
## Review: fix recovery rollback handling

Overall this is solid defensive work — the rollback-before-mutation pattern, lock refactoring to avoid deadlocks, `isMountedRef` guards, and test coverage are all well done. One genuine bug found and fixed (commit pushed to this branch).

### Bug fixed: `restoreSnapshotBestEffort` can leave mnemonic/secret mismatched

`restoreSnapshotBestEffort` writes mnemonic and secret back independently. If the mnemonic rollback fails but the secret rollback succeeds, storage ends up with the **new mnemonic** paired with the **old secret**. Deriving from that mnemonic produces a different key than what's stored — silent data corruption that could lock the user out of recovery.

**Reproduction scenario:**
1. `restoreSecretFromMnemonic` writes new mnemonic (succeeds), then new secret (fails)
2. Rollback: write old mnemonic back (fails), write old secret back (succeeds)
3. Result: new mnemonic + old secret = mismatched pair

**Fix pushed in `f5a29eae`:** When any rollback write fails (partial rollback), clear both keys. A missing pair is recoverable (`ensureSecret` regenerates from scratch); a mismatched pair is silently corrupt.

The commit includes a test that proves the mismatch scenario and verifies both keys are cleared.

### Everything else looks good

- Lock refactoring (`readStoredSecretSnapshotUnlocked` for use inside `withSecretLock`) correctly avoids deadlock
- `isMountedRef` navigation guards are proper React pattern
- TourScreen try/catch with KYC fallback is sensible
- Test coverage is thorough for the failure/rollback scenarios

Approving with the fix already applied. ✅

PR #1906, opened 2026-04-02
Harden native lifecycle result handling

transphorm:
## Summary

- Prevent duplicate result emissions by introducing a `LifecycleResultGate` (Android) and guarding `setResult`/`dismiss` with `hasEmittedResult` checks (iOS) so only the first caller wins
- Change success validation from `!= false` to `== true` on both platforms, requiring an explicit `success: true` boolean rather than treating missing/non-false values as success
- Extract `LifecycleResultEnvelope` on both platforms to unwrap nested `result` objects from the bridge params before forwarding to callbacks
- Add `onFailure` callback on iOS to route failed verification results separately from successful ones, with a `SelfLifecycleResultError` that extracts error messages from the payload

### Native Shell — Android

- `LifecycleHandler.kt`: replace `hasResult` boolean with `LifecycleResultGate` that prevents both duplicate `setResult` and `dismiss` after `setResult`; use `LifecycleResultEnvelope` to unwrap nested result and extract success strictly as `== true`
- `LifecycleResultGateTest.kt`: tests for first-claim-wins, dismiss-then-setResult, setResult-then-dismiss, and repeated claims
- `LifecycleResultEnvelopeTest.kt`: tests for nested unwrap, fallback to raw params, and strict success extraction

### Native Shell — iOS

- `LifecycleHandler.swift`: add `onFailure` callback; guard `setResult` and `dismiss` to emit only once via `hasEmittedResult`; use `SelfLifecycleResultEnvelope` for payload extraction and success check; route failures to `onFailure` with `SelfLifecycleResultError`
- `LifecycleHandlerRaceTests.swift`: tests for setResult→dismiss, dismiss→setResult, failure routing, double setResult, and double dismiss
- `LifecycleResultEnvelopeTests.swift`: tests for payload extraction, success extraction, nil handling, and error message extraction

## Test Plan

- [ ] `cd packages/native-shell-android && ./gradlew test` passes (LifecycleResultGate + Envelope tests)
- [ ] `cd packages/native-shell-ios && swift test` passes (LifecycleHandler race + Envelope tests)
- [ ] Manual: WebView calls `setResult({success: true})` then `dismiss` — consumer receives result, not cancellation
- [ ] Manual: WebView calls `dismiss` before `setResult` — consumer receives cancellation, result is ignored
- [ ] Manual: WebView calls `setResult({success: false, error: {message: "..."}})` — iOS `onFailure` fires with error

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate result emissions across Android and iOS lifecycle flows; dismissal and result reporting now reliably occur only once.
  * Improved failure handling so failed verifications invoke explicit failure callbacks and success is recognized only for an explicit boolean true.
  * Ensured dismissals route callbacks consistently whether a view controller is present or not.

* **Tests**
  * Added extensive tests covering result extraction, success/failure detection, gating behavior, and race conditions on both platforms.


Tranquil-Flow (commented):
## Review: harden native lifecycle result handling

Thorough review — no issues found. This is solid work.

### What was verified

**`== true` strict success check:** Traced through the full bridge flow (TypeScript `VerificationResult` → bridge serialization → native deserialization → handler). The TS layer always sends `success: true` explicitly. The stricter check is fail-closed, which is correct for a verification system. Safe change.

**`LifecycleResultGate` thread safety:** All `tryClaim()` calls on Android are serialized on the main thread via `activity.runOnUiThread`. On iOS, both `setResult` and `dismiss` run inside `MainActor.run`. No concurrent access is possible. The plain boolean is fine.

**iOS weak-to-strong callback:** No retain cycle — the callback doesn't own the handler, and the handler holds only a weak ref to the VC. The old `weak var callback` was actually the bug — it could silently drop results if the VC was deallocated during dismiss animation. The strong capture is the correct fix.

**First-caller-wins pattern:** Tested all real-world sequences — setResult→dismiss, dismiss→setResult, double-setResult, double-dismiss, and iOS VC/no-VC paths. All correctly guarded.

### Minor note

The `LifecycleResultEnvelope.extractPayload` nested `result` unwrap path appears unreachable from the current WebView bridge code (the bridge sends `VerificationResult` flat as params, not nested). The fallback handles this correctly — just confirming this is intentional defensive coding.

Looks good to merge. ✅

PR #1907, opened 2026-04-02
Add remote webview integrity checks

transphorm:
## Summary

- Add SHA-256 integrity-verified remote WebView content loading to both Android and iOS native shells, enabling SDK consumers to optionally serve the web app from a remote URL while maintaining security through hash verification
- Bundled assets remain the default and load first; remote content only replaces the WebView if the fetched HTML passes integrity checks (correct HTTP status, acceptable content type, matching SHA-256 hash)
- Refactor `AndroidWebViewHost` and `SelfWebViewHost` (iOS) to use origin-matching helpers and extract magic strings into constants
- Add `SelfBundledAssetSchemeHandler` on iOS to serve bundled assets via a custom `self-sdk://` URL scheme instead of relying on the Vercel URL

### Native Shell — Android
- `SelfSdkConfig` / `SelfSdk` / `SelfVerificationActivity`: thread new `remoteWebAppBaseUrl` and `remoteWebAppIntegritySha256` config fields through to the WebView host
- `AndroidWebViewHost`: refactor URL/origin handling into helper methods (`isBundledOrigin`, `isDebugOrigin`, `isAllowedRemoteOrigin`, `resolvePort`); add `maybeLoadVerifiedRemoteContent` that fetches, hashes, and conditionally loads remote HTML; normalize path handling in asset loader
- `RemoteContentIntegrity`: new utility object for SHA-256 hashing, hash normalization, and content-type validation
- `RemoteContentIntegrityTest`: unit tests for hash normalization, SHA-256 computation, and content-type acceptance

### Native Shell — iOS
- `SelfSdkConfig` / `SelfSdk`: add `remoteWebAppBaseURL` and `remoteWebAppIntegritySha256` config fields; pass `onFailure` callback
- `SelfWebViewHost`: add `WKNavigationDelegate` for navigation policy; add `loadVerifiedRemoteContent` with `URLSession`-based fetch + `CryptoKit` SHA-256 verification; add `SelfBundledAssetSchemeHandler` for custom `self-sdk://app` scheme; refactor URL construction into `makeEntryURL` / `makeBundledEntryURL`
- `RemoteContentIntegrity`: Swift equivalent of the Android utility
- `RemoteContentIntegrityTests`: XCTest suite for normalization and mime-type checks
- `Package.swift`: add test target

### Config
- `build.gradle.kts`: add JUnit and kotlin-test dependencies

## Test Plan

- [ ] `cd packages/native-shell-android && ./gradlew test` passes (RemoteContentIntegrity tests)
- [ ] `cd packages/native-shell-ios && swift test` passes (RemoteContentIntegrity tests)
- [ ] SDK launches with no remote config — loads bundled assets as before (no regression)
- [ ] SDK launches with valid `remoteWebAppBaseUrl` + correct `remoteWebAppIntegritySha256` — remote content loads
- [ ] SDK launches with valid URL but wrong hash — falls back to bundled content
- [ ] SDK launches with unreachable URL — falls back to bundled content

## Summary by CodeRabbit

* **New Features**
  * Added optional config parameters for a remote web app base URL and its SHA-256 integrity value (Android & iOS).
  * Web view hosts now optionally fetch, verify (SHA-256 + content type), and load remote HTML when configured.
  * Tightened navigation and permission restrictions to only allow bundled, debug, or the configured remote origins.

* **Tests**
  * Added unit tests for integrity normalization, hash computation, and content-type acceptance.

### DIFF
--- a/packages/native-shell-android/src/main/kotlin/xyz/self/sdk/api/SelfSdk.kt
+++ b/packages/native-shell-android/src/main/kotlin/xyz/self/sdk/api/SelfSdk.kt
@@ -35,6 +35,10 @@ object SelfSdk {
                 config.chainID?.let { putExtra(SelfVerificationActivity.EXTRA_CHAIN_ID, it) }
                 config.userDefinedData?.let { putExtra(SelfVerificationActivity.EXTRA_USER_DEFINED_DATA, it) }
                 config.selfDefinedData?.let { putExtra(SelfVerificationActivity.EXTRA_SELF_DEFINED_DATA, it) }
+                config.remoteWebAppBaseUrl?.let { putExtra(SelfVerificationActivity.EXTRA_REMOTE_WEB_APP_BASE_URL, it) }
+                config.remoteWebAppIntegritySha256?.let {
+                    putExtra(SelfVerificationActivity.EXTRA_REMOTE_WEB_APP_INTEGRITY_SHA256, it)
+                }
             }
         activity.startActivityForResult(intent, requestCode)
     }
--- a/packages/native-shell-android/src/main/kotlin/xyz/self/sdk/api/SelfSdkConfig.kt
+++ b/packages/native-shell-android/src/main/kotlin/xyz/self/sdk/api/SelfSdkConfig.kt
@@ -19,6 +19,8 @@ data class SelfSdkConfig(
     val chainID: Int? = null,
     val userDefinedData: String? = null,
     val selfDefinedData: String? = null,
+    val remoteWebAppBaseUrl: String? = null,
+    val remoteWebAppIntegritySha256: String? = null,
 )
 
 class SelfSdkException(
--- a/packages/native-shell-android/src/main/kotlin/xyz/self/sdk/webview/AndroidWebViewHost.kt
+++ b/packages/native-shell-android/src/main/kotlin/xyz/self/sdk/webview/AndroidWebViewHost.kt
@@ -281,7 +281,20 @@ class AndroidWebViewHost(
                 Log.w("WebViewHost", "Remote web app integrity check failed due to unexpected content type ${connection.contentType}")
                 null
             } else {
-                val body = connection.inputStream.use { it.readBytes() }
+                val body = connection.inputStream.use { stream ->
+                    val buffer = java.io.ByteArrayOutputStream()
+                    val chunk = ByteArray(8192)
+                    var totalRead = 0
+                    var bytesRead: Int
+                    while (stream.read(chunk).also { bytesRead = it } != -1) {
+                        totalRead += bytesRead
+                        if (totalRead > MAX_REMOTE_ENTRY_BYTES) {
+                            throw IllegalStateException("Remote entry response exceeded ${MAX_REMOTE_ENTRY_BYTES} bytes")
+                        }
+                        buffer.write(chunk, 0, bytesRead)
+                    }
+                    buffer.toByteArray()
+                }
                 if (sha256Hex(body) == normalizeSha256(expectedSha256)) {
                     String(body, Charsets.UTF_8)
                 } else {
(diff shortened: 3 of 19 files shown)


Tranquil-Flow (commented):
## Review: add remote webview integrity checks

Good security work — the SHA-256 integrity gating, HTTPS-only enforcement, origin-locked navigation, and path traversal protection are all correctly implemented. One real security gap worth discussing.

### Sub-resource integrity gap

The SHA-256 hash only covers the entry HTML document. JavaScript, CSS, and other resources loaded by that HTML are fetched from the remote server **without any integrity check**. The navigation policy allows same-origin requests, so they load and execute.

An attacker who compromises the CDN could keep the HTML byte-for-byte identical (passes the hash) and swap out a JS bundle. The malicious JS loads and executes in the WebView context.

**Possible mitigations:**
- Require the entry HTML to use [Subresource Integrity](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) attributes (`integrity` on `<script>` and `<link>` tags) — the browser enforces these natively
- Extend the hash manifest to cover sub-resources (more complex, probably overkill)
- Document the expectation that the entry HTML must be a self-contained SPA with all JS inlined (if that's the intent)

This isn't necessarily a blocker — depends on the threat model and whether the entry HTML is self-contained. But worth an explicit decision.

### Previously flagged items that are NOT issues

- **iOS redirect following:** Not exploitable — the hash check runs on whatever bytes arrive regardless of redirects. SHA-256 collision is cryptographically infeasible.
- **iOS MIME case sensitivity:** `URLSession` normalizes MIME types in practice. Non-issue.
- **Path traversal on Android:** Android `AssetManager` is sandboxed — `..` doesn't escape the assets directory. iOS correctly needs the explicit protection since it uses file:// URLs.

### Also needs attention

This PR has **merge conflicts** with dev that need resolving before it can merge.

Good feature, worth merging after the conflict resolution and a decision on the sub-resource question. 👍

Tranquil-Flow:
## Fix pushed: Subresource Integrity (SRI) for build output

Pushed commit `64aead661` to close the sub-resource integrity gap flagged in the earlier review.

### The problem

The SHA-256 remote integrity check only covers the entry HTML (428 bytes). The actual app code — a 7MB JS bundle and CSS — loads as separate files with no integrity verification. A compromised CDN could keep the HTML identical (passes the hash) and swap the JS bundle.

### The fix

Added a custom Vite plugin (`subresourceIntegrity`) in `vite.config.ts` that runs after all files are written to disk (`closeBundle` hook) and injects `integrity="sha384-..."` attributes into all `<script>` and `<link>` tags in `index.html`.

Before:
```html
<script type="module" crossorigin src="/assets/index-B_6L0fdQ.js"></script>
<link rel="stylesheet" crossorigin href="/assets/index-CsEQqL_b.css">
```

After:
```html
<script type="module" crossorigin src="/assets/index-BuePNTU_.js" integrity="sha384-JdRkqyMOCounbbr2Dd7WoAwPVg1BMmys8tY0WGv4Dp08F0Yz3zFbG/wi7ezhpI0e"></script>
<link rel="stylesheet" crossorigin href="/assets/index-CsEQqL_b.css" integrity="sha384-LxFmkx+E0ynLZNNUZjPUu2X45Jn4M88Nz9oVn9M1v2nrabIQCopbDbLnjaWcfCOe">
```

The browser natively enforces SRI — if the fetched JS/CSS doesn't match the hash, the browser blocks it entirely. No native code changes needed.

### Tests

Added `src/test/sri.test.ts` with 4 tests that verify:
1. Script tags have `integrity` attributes
2. Stylesheet links have `integrity` attributes
3. Script hashes match actual file contents on disk
4. Stylesheet hashes match actual file contents on disk

All 48 webview-app tests pass.

### Why `closeBundle` instead of `generateBundle`

The Rollup `generateBundle` hook runs before sourcemap comments (`//# sourceMappingURL=...`) are appended to JS files. Hashing at that stage produces mismatched hashes. `closeBundle` runs after all files are written to disk, so hashes match the final bytes the browser will fetch.

### Note

This PR still has merge conflicts with `dev` that need resolving (from #1905 and #1906 merging).

PR #2108, opened 2026-05-25
spike(rn-sdk): KMP-backed secureStorage bridge (Path A)

remicolin:
## Summary

Prototype that proves `rn-sdk` can wrap `kmp-sdk` for the `secureStorage` domain on Android, behind a feature flag. The flag is off by default — existing `SelfVerification` behavior is unchanged for every consumer.

- Spec (one-page): [`PATH-A-rn-wraps-kmp.html`](https://github.com/selfxyz/self/blob/spike/rn-wraps-kmp/specs/projects/sdk/workstreams/webview-in-app/plans/PATH-A-rn-wraps-kmp.html)
- Research notes: [`SPIKE-rn-wraps-kmp.md`](https://github.com/selfxyz/self/blob/spike/rn-wraps-kmp/specs/projects/sdk/workstreams/webview-in-app/plans/SPIKE-rn-wraps-kmp.md)

## What it does

When `<SelfVerification useKmpBridge>` is `true`, bridge messages with `domain === "secureStorage"` are forwarded to a new `SelfBridgeModule` (Kotlin) that drives `kmp-sdk`'s `MessageRouter` + `SecureStorageBridgeHandler` + `EncryptedSharedPreferencesProvider`. Responses flow back via `DeviceEventEmitter` → `KmpBridgeTransport` → `webView.injectJavaScript`. All other domains keep using the existing TS handlers.

## What it proves

Validation results from this branch:

- ✅ `kmp-sdk` publishes a consumable AAR via `publishToMavenLocal`
- ✅ `packages/rn-sdk/android/` resolves `xyz.self.sdk:shared-android:0.1.0` from `mavenLocal()`
- ✅ `rn-sdk-test-app/android` builds with the KMP AAR resolved through `allprojects { mavenLocal() }`
- ✅ Kotlin 1.9.24 (rn-sdk) consuming Kotlin 2.1.0-compiled KMP classes compiles cleanly
- ✅ rn-sdk TS typecheck + vitest pass (144 tests, including 4 new for `KmpBridgeTransport`)
- ✅ Native glue total: ~50 LOC in `SelfBridgeModule.kt` (abort condition was 250)

## What's not validated yet

- Runtime smoke (set / get / remove round-trip on a real Android device) — requires a developer with a device to open the `KmpBridgeSmokeScreen` in `rn-sdk-test-app` and exercise it. The screen is wired and ready.
- iOS — explicitly out of scope. KMP iOS still has cinterop disabled.
- Self Wallet integration — explicitly out of scope.

## Open risks (from the spec)

- **Two-repo split.** `selfxyz/self-webview-sdk` holds a forked `kmp-sdk` where MOD-01/02/04 are landing. This PR consumes _this_ repo's `packages/kmp-sdk/`. Convergence is a separate decision and not blocking this PR.
- **mavenLocal consumer leak.** `rn-sdk-test-app/android/build.gradle` adds `mavenLocal()` so the AAR resolves. Real consumers (Self Wallet) would need the same — or wait for SD-06 (`SELF-2534`) to publish to hosted Maven.
- **`maven-publish` Gradle 8.13 task-dep bug in iOS sources jar** — `./gradlew :shared:publishToMavenLocal` fails on the iOS publication side; `:shared:publishAndroidReleasePublicationToMavenLocal :shared:publishKotlinMultiplatformPublicationToMavenLocal` works. Pre-existing issue in `kmp-sdk/shared/build.gradle.kts`, not blocking the prototype.

## Test plan

- [ ] `cd packages/kmp-sdk && ANDROID_HOME=… ./gradlew :shared:publishAndroidReleasePublicationToMavenLocal :shared:publishKotlinMultiplatformPublicationToMavenLocal` succeeds
- [ ] `cd packages/rn-sdk && yarn typecheck && yarn vitest run && yarn build` all pass
- [ ] `cd packages/rn-sdk-test-app/android && ./gradlew :selfxyz_rn-sdk:assembleRelease` builds the rn-sdk lib consuming the KMP AAR
- [ ] On an Android device or emulator: open `KmpBridgeSmokeScreen` in `rn-sdk-test-app`, run set/get/remove round-trip, observe log entries
- [ ] Default behavior of `<SelfVerification>` (without `useKmpBridge`) regresses nothing for any existing consumer

## What this PR is not

Not a production rollout. Merging this PR is the explicit "Phase 2 prototype lands; we decide on Path A vs B." A follow-up workstream would retire the duplicated TS handlers if this approach is greenlit.


remicolin:
## Smoke test result: **FAIL** (Android, emulator)

Device: Android emulator `qa(AVD)` running Android 14 (API 34, `emulator-5554`), `1080x2400@440dpi`.

### What I saw

The smoke screen renders fine and accepts taps. The four taps `set / get / remove / get` each fire a `dispatch` log entry on screen, **but no `response` ever comes back** — the promises hang.

On-screen log after running the sequence:

```
03:10:22.275 dispatch set    req-1779…
03:10:26.266 dispatch get    req-1779…
03:10:30.389 dispatch remove req-1779…
03:10:34.452 dispatch get    req-1779…
(no response entries — log ends here)
```

Expected per the spike doc:

```
set    -> response success=true (data null)
get    -> response success=true, data { value: "hello" }
remove -> response success=true (data null)
get    -> response success=true, data { value: null }
```

### Diagnostic notes

- The "SelfBridge native module not linked" warning **did not** fire, and the JS `dispatch` path was reached, so autolinking + module discovery is working: `NativeModules.SelfBridge` is defined and `transport.isAvailable()` returned `true`.
- The `transport.dispatch(payload)` call did not throw (no `reject` log entry would otherwise appear).
- `adb logcat` shows nothing at all from the app PID for `SelfBridge*`, `SelfCrypto*`, `MessageRouter`, `SecureStorage*`, or `EncryptedSharedPreferences*` during the tap window. That's expected — `SelfBridgeModule.kt` and `SelfCryptoPackage.kt` don't emit any `Log.d`/`Log.i` of their own, so the native path is opaque from the outside.
- No JS warnings/errors in logcat for the app PID during the dispatch window either (other than unrelated `unknown:ReactNative` "Attempt to set local data for view with unknown tag: -1" warnings that fire during initial Yoga layout).

This matches the **"Promise hangs (no response logged) — DeviceEventEmitter wiring issue in `SelfBridgeModule.kt` or `KmpBridgeTransport.ts`"** failure pattern called out in the spike handoff.

### Likely directions for the author (not fixed by me, per scope)

1. `SelfBridgeModule.routeMessage` runs `router.onMessageReceived(rawJson, isTrustedSource = true)` synchronously, but `MessageRouter` dispatches handler work onto its own `CoroutineScope`. Worth confirming the handler actually invokes `sendToWebView` for `secureStorage.{set,get,remove}` — if `EncryptedSharedPreferencesProvider.set/get/remove` throws or suspends without resolving, the response is never produced.
2. `emitInjection` calls `getJSModule(RCTDeviceEventEmitter::class.java).emit(...)` from the coroutine `Dispatchers.Default` thread. Some RN versions complain when emitting before catalyst init or off the JS thread; check there isn't a silent swallow inside `BridgelessReactContext.emit`.
3. Add temporary `Log.d("SelfBridge", ...)` lines at: entry to `routeMessage`, just before `emitInjection`, and inside the `MessageRouter.sendToWebView` callback — that alone will localize whether the hang is in handler execution or in the emitter wiring.

### Repro state

- Branch: `spike/rn-wraps-kmp` @ `3b9f0eb0`
- Local `mavenLocal` populated via the two-task workaround (full `publishToMavenLocal` still fails on the iOS sources-jar task-dep bug).
- `rn-sdk` dist was stale on first run — needed `yarn build` in `packages/rn-sdk/` before `rn-sdk-test-app` could resolve `KmpBridgeTransport`. Not a runtime issue, but worth a note in the spike doc so the next runner doesn't lose 10 minutes on a "Cannot read property 'prototype' of undefined" red box.

Video of the run + screenshots captured locally; handing off to @tappy for delivery.

remicolin:
## Smoke retest after fix: **PASS** (Android emulator)

Pushed [`12d007be`](https://github.com/selfxyz/self/commit/12d007be) — Metro `blockList` now also excludes `packages/rn-sdk/node_modules/{react,react-native}` and `packages/mobile-sdk-alpha/node_modules/{react,react-native}`, matching the existing pattern for `app/` and `mobile-sdk-demo/`.

### Root cause

Two `react-native` copies were ending up in the Metro bundle:
- `rn-sdk-test-app/node_modules/react-native` (used when bundling test-app sources)
- `rn-sdk/node_modules/react-native` (used when bundling code from `@selfxyz/rn-sdk`)

Both copies eagerly executed `BatchedBridge.js`, each creating its own `MessageQueue` singleton and each overwriting `global.__fbBatchedBridge` via `Object.defineProperty`. Last-write-wins, so the final `__fbBatchedBridge` was the one from `rn-sdk`'s RN copy — which only had `RNCWebViewMessagingModule` registered as a callable JS module. `RCTDeviceEventEmitter`, `Systrace`, `JSTimers`, etc. were registered on the *other* singleton from `setUpBatchedBridge.js`.

Native `getJSModule(RCTDeviceEventEmitter::class.java).emit(...)` routed through `mCatalystInstance.callFunction("RCTDeviceEventEmitter", "emit", …)` over the JNI bridge → arrived at the wrong `__fbBatchedBridge` → no callable module under that name → call silently no-oped. Native side showed `emit returned` happily; JS listener never fired.

Verifiable from the bundle size: pre-fix bundle was 11.3 MB, post-fix is 5.5 MB. After the fix, `__fbBatchedBridge._lazyCallableModules` contains the full set: `Systrace,JSTimers,HeapCapture,SamplingProfiler,RCTLog,RCTDeviceEventEmitter,RCTNativeAppEventEmitter,GlobalPerformanceLogger,HMRClient,AppRegistry,RCTEventEmitter,RNCWebViewMessagingModule`.

### Smoke result

On-screen log (timings are dispatch → response, full round-trip):

```
06:07:29.593 dispatch set    req-1779…
06:07:29.804 response req-1779 success=true 210ms
06:07:33.671 dispatch get    req-1779…
06:07:33.725 response req-1779 success=true  53ms
06:07:36.760 dispatch remove req-1779…
06:07:36.793 response req-1779 success=true  33ms
06:07:39.799 dispatch get    req-1779…
06:07:39.819 response req-1779 success=true  20ms
```

No `SelfBridge native module not linked` warning, no crash, no hung request. Set/get/remove/get round-trips behave exactly as the spike's success criteria specifies.

### Device

Android 14 emulator (`qa(AVD)`, `emulator-5554`), 1080×2400@440dpi, `newArchEnabled=false`, `hermesEnabled=true`.

### Notes for the spike doc

- Add the Metro dedup gotcha to `specs/projects/sdk/workstreams/webview-in-app/plans/SPIKE-rn-wraps-kmp.md` and `PATH-A-rn-wraps-kmp.html` — any consumer app that uses `@selfxyz/rn-sdk` in a workspace with multiple `react-native` installs will hit the same silent-drop, and the failure mode (promises hang, no native logs because handlers don't have logging) is the same one I tripped on for ~30 minutes.
- The fix is consumer-side (in the app's `metro.config.cjs`), not in `rn-sdk` itself. We could also tighten `rn-sdk/package.json` to declare `react-native` only as a peerDependency, which would prevent yarn from installing a sibling copy, but that's a follow-up.

Branch is now ready to merge into `feat/webview-in-app`.

PR #2217, opened 2026-07-16
add e-Aadhaar PDF upload with password entry

seshanthS:
## Summary

- Adds a PDF path to the Aadhaar registration flow: users can upload their password-protected e-Aadhaar PDF, enter its password, and register — alongside the existing QR-image upload.
- Native modules (iOS PDFKit, Android) decrypt the PDF, rasterize the relevant pages, and locate/decode the Secure QR, returning the same QR payload as the photo-library flow.
- The password hint copy is served via Remote Config (`aadhaar_pdf_password_hint`) since UIDAI may change the password format; an empty remote value hides the hint.
- UI fixes along the way: the Aadhaar navbar progress bar now adapts to the 3-step PDF path vs the 2-step QR path, bottom padding uses real safe-area insets (fixes overlap with the Android gesture pill), and only the tapped upload button shows "Processing...".

### React Native app

- New `AadhaarPdfPasswordScreen` with uncontrolled password input, incorrect-password retry, remote-configurable hint, and best-effort cleanup of the cached decrypted PDF copy on exit.
- `AadhaarUploadScreen`: adds the "Upload e-Aadhaar PDF" action via `expo-document-picker`; per-action processing state so the QR and PDF buttons no longer share a "Processing..." label.
- `AadhaarNavBar`: progress bar renders 2 or 3 segments depending on the flow (detects the PDF path from the nav stack on success) with cumulative fill.
- Both Aadhaar screens use `useSafeAreaInsets().bottom` for bottom padding instead of the window-height heuristic.
- `qrScanner` integration exposes `scanQRCodeFromPDF(fileUri, password)` and availability checks; navigation types/routes wired for the new screen.
- Remote Config: string flags only override baked-in defaults when the value source is `remote`, so an explicit empty string is honored.

### Native iOS

- `QRScannerBridge.scanQRCodeFromPDF`: PDFKit-based decrypt (rejects `INVALID_PASSWORD` on wrong password), full-page decode with a high-DPI crop fallback of the known Secure-QR regions (first/last page).
- Photo-library QR decoding refactored into a shared `QRImageDecoder` used by both flows.

### Native Android

- `PdfQrHelper` + `QRCodeScannerModule.scanQRCodeFromPDF` mirroring the iOS behavior; `QrCodeDetectorProcessor` updated for shared decoding.

### Tests / config

- Navigation test covers the new route; remote-config tests cover the remote-vs-default string flag behavior; jest mock for `expo-document-picker`.
- `expo-document-picker` added (package.json, pnpm-lock, Podfile.lock). The pnpm-lock diff is large because pnpm 11.12.0 re-resolves peer deps (`supports-color` threading) on any install; regenerated with `pnpm install --lockfile-only`.

## Test Plan

- [ ] `pnpm lint && pnpm types` passes
- [ ] `cd app && pnpm test` passes
- [ ] Manual: upload a password-protected e-Aadhaar PDF on iOS and Android — wrong password shows inline error, correct password registers
- [ ] Manual: QR-image upload path still works; only the tapped button shows "Processing..."
- [ ] Manual: progress bar shows 2 segments on the QR path, 3 on the PDF path; bottom button clears the Android gesture pill

## Summary by CodeRabbit

* **New Features**
  * Added support for uploading password-protected e-Aadhaar PDFs and scanning the embedded secure QR.
  * Introduced a dedicated PDF password entry screen with hint text, processing state, and retry-friendly handling.
  * Added a new PDF/password scan step to navigation and updated the multi-step progress UI.
* **Bug Fixes**
  * Improved QR extraction from PDFs and standardized outcomes for “QR not found” and “incorrect password.”
  * Updated password hint flag behavior so remote values (including intentionally empty hints) are respected.
* **Tests**
  * Updated Jest coverage for navigation, remote hint flag behavior, and document picker mocking.


seshanthS:
Addressed the review findings in 1d62148ef:

- **Fail closed on QR selection (CodeRabbit Major / Codex P1, both platforms):** all native decode paths now only accept the Aadhaar Secure QR shape (numeric, ≥100 chars). Android's `values.firstOrNull()` fallback is removed and the ZXing tile path validates its result; iOS decodes all features per pass and filters, so an unrelated page QR no longer short-circuits the tile/crop retries. The photo-library flow's behavior is unchanged.
- **Android memory (CodeRabbit Major / Codex P2):** page rasterization is clamped to a ~25 MP pixel budget (A4 @ 500 DPI unchanged; larger pages render proportionally lower), tile allocation skips on `OutOfMemoryError`, and `scanQRCodeFromPDF` catches `Throwable` so the JS promise always settles.
- **Cached PDF lifetime (Codex P2):** the cache copy is deleted as soon as the native scan settles (kept only across wrong-password retries); the unmount cleanup remains as a backstop.
- **`SCAN_SUCCEEDED` accuracy (Codex P2):** tracking moved to the `PROVING_AADHAAR_UPLOAD_SUCCESS` listener in `selfClientProvider`; `processAadhaarQRCode` resolves even on failure, so the screens no longer track success themselves.
- **Error-screen retry (Codex P2):** "Try upload again" now pops to `AadhaarUpload` (via `StackActions.popTo`, preserving params) instead of returning to the password screen with the same unusable PDF.

PR #2235, opened 2026-07-24
fix(gate): Google USAT — route to selector for eligible alternatives + endpoint-agnostic match (SELF-3667)

seshanthS:
Two related fixes to the Google USAT faucet verification gate.

## 1. Route to the document selector when an eligible alternative exists
**Bug:** A user with a KYC document *selected* but an eligible document (e.g. Aadhaar) *also registered* was shown the "High-security ID required" block modal when the faucet was opened via **deeplink** (and, identically, via **QR scan** and **earn-points**).

**Cause:** the three entry points pre-checked the gate with `evaluateGoogleUsatGate`, which only inspects the *selected* document, and hard-blocked — even though `ProvingScreenRouter` (where they all navigate) already knows how to steer users with an eligible alternative to the document selector.

**Fix:** new `evaluateGoogleUsatEntryGate` helper that blocks only when there is *no* eligible document at all; otherwise it defers to `ProvingScreenRouter`, which opens the selector. The three entry points now use it. `evaluateGoogleUsatGate` is left unchanged (its selected-doc-only contract is relied on by `ProvingScreenRouter` and covered by an existing test).

Tests: new `evaluateGoogleUsatEntryGate` unit suite; repointed deeplink + earn-points mocks; added an earn-points block test.

## 2. Temporarily match the faucet by scope + appName only (SELF-3667)
The gate identity match hard-codes the faucet `endpoint`, so an endpoint rotation silently disables the gate until an app release updates the constant. As a stopgap (remote-config is deferred, see SELF-3667), the `endpoint` clause in `findRestrictedAppPolicy` is commented out — matching on `scope` + `appName` only — so the gate survives endpoint rotation without a release.

Contained to the Google USAT gate (only policy in the registry). No security regression: all identity fields are verifier-controlled JSON, so endpoint never provided spoof resistance; the gate is a UX guard with server-side eligibility enforcement and fails open. All temporary edits are marked `TEMPORARY(SELF-3667)` for easy revert.

Follow-up to restore endpoint matching via Remote Config: **SELF-3667**.

## Validation
- `app` `tsc` clean; SDK `tsc` clean.
- `eslint` clean on all changed files.
- `jest`: googleUsatGate + deeplinks + useEarnPointsFlow + ProvingScreenRouter suites pass (92 tests).
- `vitest`: SDK restrictedApps + googleUsat suites pass (26 tests).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Google USAT entry checks across points earning, QR code, and deeplink flows.
  - Users are now allowed to continue when an eligible alternative document is available.
  - Blocked users are consistently shown the verification prompt instead of being sent into an incomplete flow.
  - Google USAT access checks remain available during faucet endpoint changes, reducing unnecessary access blocks.

- **Tests**
  - Added coverage for eligible alternatives, missing documents, blocked access, and error handling.

### DIFF
--- a/app/src/hooks/useEarnPointsFlow.ts
+++ b/app/src/hooks/useEarnPointsFlow.ts
@@ -19,7 +19,7 @@ import {
 } from '@/services/points';
 import useUserStore from '@/stores/userStore';
 import { useVerificationGateStore } from '@/stores/verificationGateStore';
-import { evaluateGoogleUsatGate } from '@/utils/googleUsatGate';
+import { evaluateGoogleUsatEntryGate } from '@/utils/googleUsatGate';
 import { registerModalCallbacks } from '@/utils/modalCallbackRegistry';
 
 type UseEarnPointsFlowParams = {
@@ -40,7 +40,7 @@ export const useEarnPointsFlow = ({
 
   const navigateToPointsProof = useCallback(async () => {
     const selfApp = await pointsSelfApp();
-    const gate = await evaluateGoogleUsatGate(selfClient, selfApp);
+    const gate = await evaluateGoogleUsatEntryGate(selfClient, selfApp);
     if (gate === 'block') {
       selfClient.trackEvent(ProofEvents.GOOGLE_USAT_BLOCKED, {
         entry_point: 'earn_points',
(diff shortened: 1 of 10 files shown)


seshanthS:
@codex

